PR #3, opened 2019-11-20
Update draft-ietf-rats-architecture.md

nedmsmith:
"added diagram for multiple attesters with different evidence formats"


dthaler:
Thanks for submitting this.  In my proposed slides for Friday, I made some additions for review.  Would like feedback on that, since we may want to update the diagram and/or figure title in this PR.

Specifically, the changes use more specific examples of the evidence formats, and also show multiple attestation result formats in the same diagram.

nedmsmith:
The revised diagram shows support for multiple Results formats excluding TPM. Is there an architectural objective to limit only TPM? It seems the motivation for limiting multiple results formats is to reduce complexity (security objective). However, if the architecture doesn't have this motivation, then filtering TPM formats seems arbitrary. Possibly, the architecture intends to allow multiple Results formats to accommodate protocol constraints. For example, TLS didn't anticipate exchanging Evidence in a handshake, hence it may be necessary to encode it in the certificate path.

dthaler:
No specific objective, if you think it's good to add, that's fine with me.

PR #6, opened 2019-11-27
Terminology, dataflow/roles, and trust model

dthaler:
Here's a first attempt to combine terminology sections, and trust model.
Includes new terms Endorser and Appraisal Policy per recent discussion.


dthaler:
> The word 'health' is generally being used where other words such as 'trustworthiness', or 'state' could be used. I'm concerned that 'health' might not easily be synonymous with 'trust'. It also seems to bias the architecture toward NEA (RFC5209) which IMO does not define attestation.

Done

dthaler:
> In the trust model section I generally don't like sentence structure that ends in a preposition. Could be worded better.

Done.  (But if you have feedback on a sentence like this in the future, please attach the comment directly to the code line under the "Files changed" tab, to make it easy to find which phrase you're referring to.)

dthaler:
> Verifier definition could be more specific about evaluating Evidence information about an Attester.

Done

dthaler:
> Attester definition seems biased towards the perspective of a Relying Party rather than being described from the perspective of a "Verifier Appraisal Policy".

Updated.  How about now?

dataflow.txt
```diff
@@ -0,0 +1,20 @@
+                              ************
+                              * Endorser *
+                              ************
```
mcr: Who is speaking?

dataflow.txt
```diff
@@ -0,0 +1,20 @@
...
+                                    |
+                  Evidence          |Endorsements
+                 Appraisal----.     |
+                  Policy      |     |             Attestation
```
mcr: Who is speaking the Appraisal Policy?

mcr: Predominantly, it is the local organization/domain-administrator.  Administrative-Domain. Or Owner.

draft-ietf-rats-architecture.md
```diff
@@ -69,27 +69,113 @@ remote attestation procedures (RATS).
...
+* Attestation Result: The evaluation results generated by a Verifier,
+    typically including information about an Attester, where the
+    Verifier vouches for the validity of the results.
+
+* Attester: An entity whose attributes must be evaluated in order to
+    determine whether the entity is considered trustworthy, such as when
+    deciding whether the entity is authorized to perform some operation.
+
+* Endorsement: A secure statement that some entity (typically a manufacturer) vouches
+    for the integrity of an Attester's signing capability.
```
William-PanWei: Does the Known-Good-Value/Reference Value fall into Endorsement or Appraisal Policy? Is it better to add some text into the explanation of the related term to explicitly express that this term includes the Reference Value?

draft-ietf-rats-architecture.md
```diff
@@ -69,27 +69,113 @@ remote attestation procedures (RATS).
...
 
-    <this section can include content from Terminology in draft-birkholz-rats-architecture
-    and draft-thaler-rats-architecture>
+This document uses the following terms:
+
+* Appraisal Policy: A set of rules that direct how a system evaluates
+    the validity of information about another entity.  Compare /security policy/
+    in {{?RFC4949}}.
+
+* Attestation: A process by which one entity (the "Attester") provides evidence
```
mcr: this term gets re-introduced here, without any annotation.
The appraisal policy are the rules on how to test the evidence. 
Here, "Attestation" is the creation of the evidence
Often people use the term "Attestation" to refer to the whole process.
Suggestion that we should define this term *somewhere*, but maybe "Remote Attestation"?

mcr: FIDO has NONE-Attestation-Type, where there is no assessment.

mcr: Let's take this to the WG, and maybe we can not get consensus on this term.

mcr: Charter says: "Remote
attestation procedures (RATS) enable relying parties to establish a level of
confidence in the trustworthiness of remote system components through the
creation of attestation evidence by remote system components and a processing
chain towards the relying party."

dataflow.txt
mcr: Endorsement is a kind of manifest, like for a ship (a shipper!)

draft-ietf-rats-architecture.md
mcr: Appraisal Policy! (No, could be both)   Reference Value is a 1:1 comparison.   But, for geo-coordinates it would be more like a policy, a calculation is required.  Appraisal Policy is a super set of Reference Value/KGV. An Endorsement is spoken by a supply chain entity, while an Appraisal Policy is spoken by an Owner.

dthaler: Appraisal Policy.   And as Michael implies, Reference Value is just an example and is not sufficient in the general case so we do not want to imply that there is necessarily such a thing as a Reference Value, except in a specific example.

dthaler:
I would like to discuss the possibility of combining the terms "Endorsement" and "Attestation Result".   Just like we have an Appraisal Policy for a verifier and an Appraisal Policy for a relying party, some might view an Attestation Result as an Endorsement for use by the relying party.   I'm ok either way, but wanted to discuss this in our next call to see if others think it would help or confuse things.

draft-ietf-rats-architecture.md
William-PanWei: Got it. Use Cases draft or 'Use Cases' section is a good place for giving an example to explain that Reference Value is one case of Appraisal Policy.

dthaler:
I still believe my original definition of Attestation is good, I haven't seen any comments here about what the problem was.  For comparison, https://en.wikipedia.org/wiki/Attestation says it's "The process of validating the integrity of a computing device such as a server needed for trusted computing.".   I'm also ok with that definition.

nedmsmith:
I’m OK with that wording for introducing the concept. It may break down if the terminology for ‘device’ and ‘trusted computing’ become overly specified / specific.

PR #15, opened 2020-01-14
Text for the Two Types of Environments section

dthaler:
Incorporating text from draft-birkholz-rats-architecture
and from my reading of the "bike shed" mailing list thread

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -273,12 +273,24 @@ plans to support in the TEEP architecture {{?I-D.ietf-teep-architecture}}.
 
 # Two Types of Environments
 
-An Attester consists of at least one Attesting Environment and Attested Environment.
-In some implementations, the Attesting and Attested Environments might be combined.
-Other implementations might have multiple Attesting and Attested Environments.
+An Attester consists of at least one Attesting Environment and at least one
+Target Environment. In some implementations, the Attesting and Target Environments
+might be combined.
+Other implementations might have multiple Attesting and Target Environments.
+One example is a set of components in a boot sequence (e.g., ROM, firmware,
+OS, and application) where a Target Environment is the
+Attesting Environment for the next environment in the boot sequence.
 
-    <this section can include Two Types of Environments content from draft-birkholz-rats-architecture
-    but can we find a better name? also this could be a subsection of something else?>
+Target Environments are measured.  They provide the raw values and
+the information to be represented in claims and ultimately expressed
+as Evidence.
+
+Attesting Environments conduct the measuring.  They collect the
+claims, format them appropriately, and typically use key material and
+cryptographic functions, such as signing or cipher algorithms, to
+create Evidence. Examples of environments that can be used as
+Attesting Environments include Trusted Execution Environments (TEE),
+embedded Secure Elements (eSE), or Hardware Security Modules (HSM).
 
 # Trust Model
 


draft-ietf-rats-architecture.md
```diff
@@ -273,12 +273,24 @@ plans to support in the TEEP architecture {{?I-D.ietf-teep-architecture}}.
...
+Target Environments are measured.  They provide the raw values and
+the information to be represented in claims and ultimately expressed
+as Evidence.
+
+Attesting Environments conduct the measuring.  They collect the
+claims, format them appropriately, and typically use key material and
+cryptographic functions, such as signing or cipher algorithms, to
+create Evidence. Examples of environments that can be used as
+Attesting Environments include Trusted Execution Environments (TEE),
+embedded Secure Elements (eSE), or Hardware Security Modules (HSM).
```
William-PanWei: Should the Trusted Platform Module (TPM) also be included here as an example?

draft-ietf-rats-architecture.md
```diff
@@ -273,12 +273,24 @@ plans to support in the TEEP architecture {{?I-D.ietf-teep-architecture}}.
...
-
-    <this section can include Two Types of Environments content from draft-birkholz-rats-architecture
-    but can we find a better name? also this could be a subsection of something else?>
+An Attester consists of at least one Attesting Environment and at least one
+Target Environment. In some implementations, the Attesting and Target Environments
+might be combined.
+Other implementations might have multiple Attesting and Target Environments.
+One example is a set of components in a boot sequence (e.g., ROM, firmware,
+OS, and application) where a Target Environment is the
+Attesting Environment for the next environment in the boot sequence.
```
William-PanWei: The carrier-grade router, one kind of composite device, is consisted of several slots, and each slot has this set of components in a boot sequence.
Calling the router "attester" and the slot "attesting environment" isn't appropriate, I think. Maybe call the router "Attester" and the slot "sub-attester"?

William-PanWei: Using "Attester" and "Sub-Attester" may give a better understanding to explain the composite device is composing of what.

draft-ietf-rats-architecture.md
```diff
@@ -273,12 +273,24 @@ plans to support in the TEEP architecture {{?I-D.ietf-teep-architecture}}.
...
 # Two Types of Environments
 
-An Attester consists of at least one Attesting Environment and Attested Environment.
-In some implementations, the Attesting and Attested Environments might be combined.
-Other implementations might have multiple Attesting and Attested Environments.
-
-    <this section can include Two Types of Environments content from draft-birkholz-rats-architecture
-    but can we find a better name? also this could be a subsection of something else?>
+An Attester consists of at least one Attesting Environment and at least one
+Target Environment. In some implementations, the Attesting and Target Environments
```
henkbirkholz: Just to make sure: Target Environment is a proposal to replace Attesting Environment (was calling it Execution Environment in the other PR before. Sry, my mistake), yes? Consequently this would result in a change of the term in the corresponding section. That is probably a separate issue?

draft-ietf-rats-architecture.md
```diff
@@ -273,12 +273,24 @@ plans to support in the TEEP architecture {{?I-D.ietf-teep-architecture}}.
 
 # Two Types of Environments
```
henkbirkholz: something like... "of an Attester" only better is required here? I am aware that the title is still a placeholder and very literal, but I had a hard time coming up with a more appropriate one. Separate issue maybe?

draft-ietf-rats-architecture.md
henkbirkholz: I think this is the "layered attestation" concept or "change of roles during the remote attestation procedure" that we should (at least I think so) dedicate another section to.

draft-ietf-rats-architecture.md
henkbirkholz: I think it is supposed to be subsumed by HSM (because I think I wrote that text, not 100% sure). I also think I am more on the neutral side to this question, as we gave up on obscure vendor neutral terminology only - as it is in the end more confusing than spelling out examples and readability is one of our main targets.

draft-ietf-rats-architecture.md
dthaler: Yes, that was the suggestion on the list.

draft-ietf-rats-architecture.md
dthaler: I agree with adding "of an Attester", good suggestion

draft-ietf-rats-architecture.md
```diff
@@ -273,12 +273,24 @@ plans to support in the TEEP architecture {{?I-D.ietf-teep-architecture}}.
...
-    but can we find a better name? also this could be a subsection of something else?>
+An Attester consists of at least one Attesting Environment and at least one
+Target Environment. In some implementations, the Attesting and Target Environments
+might be combined.
+Other implementations might have multiple Attesting and Target Environments.
+One example is a set of components in a boot sequence (e.g., ROM, firmware,
+OS, and application) where a Target Environment is the
+Attesting Environment for the next environment in the boot sequence.
+
+Target Environments are measured.  They provide the raw values and
```
mcr: Claims are collected from Target Environments.

draft-ietf-rats-architecture.md
mcr: eSE - defined by Global Platform

draft-ietf-rats-architecture.md
mcr: Eric has volunteered to write layered attestation text.

PR #27, opened 2020-02-03
Fill in conceptual messages and encoding formats text from source drafts

dthaler:

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -54,6 +54,14 @@ author:
 normative:
 
 informative:
+  OPCUA:
+    author:
+      org: OPC Foundation
+    title: "OPC Unified Architecture Specification, Part 2: Security Model, Release 1.03"
+    date: 2015-11-25
+    target: https://opcfoundation.org/developer-tools/specifications-unified-architecture/part-2-security-model/
+    seriesinfo:
+      OPC 10000-2
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -412,8 +420,119 @@ implicitly trusted is often referred to as a Root of Trust.
...
+However, while an Appraisal Policy that treats all devices from a given manufacturer the same
+may be appropriate for some use cases, it would be inappropriate to use such an Appraisal Policy
+as the sole means of authorization for use cases that wish to constrain *which* compliant devices
+are considered authorized for some purpose.  For example, an enterprise using attestation for
+Network Endpoint Assessment may not wish to let every healthy laptop from the same
+manufacturer onto the network, but instead only want to let devices that it legally owns
+onto the network.  Thus, an Endorsement may be helpful information in authenticating
+information about a device, but is not necessarily sufficient to authorize access to
+resources which may need device-specific information such as a public key for the device or
+component or user on the device.
```
mcr: It seems that the point of this paragraph is that appraisal is a key part of authorization, but often not all.

draft-ietf-rats-architecture.md
```diff
@@ -412,8 +420,119 @@ implicitly trusted is often referred to as a Root of Trust.
...
+resources which may need device-specific information such as a public key for the device or
+component or user on the device.  
+
+## Attestation Results
+
+Attestation Results may indicate compliance or non-compliance with a Verifier's Appraisal Policy.
+A result that indicates non-compliance can be used by an Attester (in the passport model) or
+a Relying Party (in the background-check model) to indicate that the Attester
+should not be treated as authorized and may be in need of remediation.  In some cases,
+it may even indicate that the Evidence itself cannot be authenticated as being correct.
```
mcr: I am still uncomfortable that in the TEEP case, the term "remediation" is used when "not having TEE thing FOO", is to download and run FOO.  This seems to diminish (trivialises) the meaning of "remediation" to me.  Do we need to define remediation?

draft-ietf-rats-architecture.md
```diff
@@ -412,8 +420,119 @@ implicitly trusted is often referred to as a Root of Trust.
...
+An Attestation Result that indicates compliance can be used by a Relying Party to make
+authorization decisions based on the Relying Party's Appraisal Policy.  The simplest such
+policy might be to simply authorize any party supplying a compliant Attestation Result
+signed by a trusted Verifier.  A more complex policy might also entail comparing information
+provided in the result against known-good reference values, or applying more complex logic
+such information. 
+
+Thus, Attestation Results often need to include detailed information about the Attester,
+for use by Relying Parties, much like physical passports and drivers licenses include
+personal information such as name and date of birth.  Unlike Evidence, which is often
```
mcr: It might be useful to distinguish information like "date of birth" into two categories.  Date of Birth can be a "known-good value", which is used to help match the person to the passport. This is used, I think, regularly by border agents to make sure that they are looking at the right person.  Date of Birth can instead be attestation results: where the relying party is a liquor sales entity.  I'm not sure what to do with this, but I'm concerned that we don't actually know in the passport analogy, which role the Date of Birth plays.

draft-ietf-rats-architecture.md
```diff
@@ -412,8 +420,119 @@ implicitly trusted is often referred to as a Root of Trust.
...
+very device- and vendor-specific, Attestation Results can be vendor-neutral if the Verifier
+has a way to generate vendor-agnostic information based on evaluating vendor-specific
+information in Evidence.  This allows a Relying Party's Appraisal Policy to be simpler,
+potentially based on standard ways of expressing the information, while still allowing
+interoperability with heterogeneous devices.
```
mcr: I think that a major motivation for the RATS work is to get to vendor-agnostic, even device-agnostic results.  So, not just "This Samsung XYZ phone is safe for HOLLYWOOD1SECURE movies", or even "This smartphone is safe for HOLLYWOOD1SECURE movies", to "This device is HOLLYWOOD1SECURE safe".  I wonder if want to say something more aspiration here.

draft-ietf-rats-architecture.md
```diff
@@ -412,8 +420,119 @@ implicitly trusted is often referred to as a Root of Trust.
...
+level of assurance against malware for example, it is important that
+information needed for evaluating the Attester be usable with existing
+protocols that have constraints around what formats they can transport.
+For example, OPC UA {{OPCUA}} (probably the most common protocol in
+industrial IoT environments) is defined to carry X.509 certificates and so
+security information must be embedded into an X.509 certificate to be passed
+in the protocol.  Thus, attestation-related information could be natively
+encoded in X.509 certificate extensions, or could be natively encoded in
+some other format (e.g., a CWT) which in turn is then encoded in an X.509
+certificate extension.
```
mcr: Who/where are we going to define these extensions?  Is it the job of RATS?   Does EAT do the work?  Or OPCUA?  Or?

draft-ietf-rats-architecture.md
```diff
@@ -412,8 +420,119 @@ implicitly trusted is often referred to as a Root of Trust.
...
+Especially for constrained nodes, however, there is a desire to minimize
+the amount of parsing code needed in a Relying Party, in order to both
+minimize footprint and to minimize the attack surface area.  So while
+it would be possible to embed a CWT inside a JWT, or a JWT inside an
+X.509 extension, etc., there is a desire to encode the information
+natively in the format that is natural for the Relying Party.
+
+This motivates having a common "information model" that describes
+the set of attestation related information in an encoding-agnostic
+way, and allowing multiple encoding formats (CWT, JWT, X.509, etc.)
```
mcr: the ,etc. concerns me. I am not opposed to making it open ended, but I'm concerned about being too flexible.
I think that what I'm saying is that if someone wants to encode to a new format, that they would be encouraged to do so as an IETF Action.  That their document would Update: this architecture document perhaps.  Not because they have to, but because it's good to bring them into the tent.  This is an aspiration thought.

mcr: what is the relationship to pre-existing vendor formats?  This could be what the "etc" refers to.

PR #28, opened 2020-02-04
Verifier also needs to be trusted by Endorser and Verifier Owner

William-PanWei:
For the situation that the Lead Attester has an internal Verifier, the Lead Attester may need to be evaluated first before it receives the Endorsements and Appraisal Policy of other Attesters. This is also a security enhancement.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -160,7 +160,7 @@ Attester, each slot is an Attester, and the main slot is the Lead Attester.
 
 Another example is a multi-chassis router composed of multiple single carrier-grade routers.
 The multi-chassis router provides higher throughput by interconnecting
-multiple routers and simpler management by being logically treated as one router.
+multiple routers and can be logically treated as one router for simpler management.
 Among these routers, there is only one main router that connects to the Verifier.
 Other routers are only connected to the main router by the network cables,
 and therefore they are managed and verified via this main router.
@@ -210,7 +210,7 @@ Verifier.
 One situation is that the Attesting Environment has no internal Verifier.
 In this situation, the Lead Attesting Environment simply combines the various
 Evidences into the final Evidence that is sent off to the remote Verifier,
-which evaluates the trusworthiness of the Composite Device,
+which evaluates the trustworthiness of the Composite Device,
 including the Lead Attester's and other Attesters', trustworthiness.
 
 The other situation is that the Lead Attesting Environment has an internal Verifier.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -212,6 +212,9 @@ Then the Claims Collector combines the Attestation Results into
 the final Evidence of the whole Composite Attester which is sent off to the remote
 Verifier, which might treat the claims obtained from the local Attestation Results
 as if they were Evidence.
+Before receiving the Endorsements and Appraisal Policies for other Attesters,
+to increase the security, the Lead Attester may first generate Evidence about
```
dthaler: This sentence is confusing.  How exactly does it "increase the security" to do things in that order?   And if the other Attesters push their evidence to the lead attester, then the order can't be guaranteed anyway, right?

William-PanWei: In the normal Attester case, the Endorser and Verifier Owner needs to trust the Verifier before sending Endorsement and Appraisal Policy to the Verifier. So it is the same in the case that the Lead Attester has an internal Verifier. "increase the security" may not be appropriate, what I want to express is to establish the trust.

The Lead Attester collecting Evidence of other Attesters happens inside the Composite Attester, so I think it doesn't matter when to collect the Evidence.

draft-ietf-rats-architecture.md
```diff
@@ -327,7 +327,11 @@ same way as any other Verifier), for evaluating these Attesters' trustworthiness
 Then the Lead Attesting Environment combines the Attestation Results into
 the final Evidence of the whole Composite Attester which is sent off to the remote
 Verifier, which might treat the claims obtained from the local Attestation Results
-as if they were Evidence.
+as if they were Evidence. In this situation, the internal Verifier may need to be
```
mcr: change internal -> local.

draft-ietf-rats-architecture.md
```diff
@@ -327,7 +327,11 @@ same way as any other Verifier), for evaluating these Attesters' trustworthiness
...
+trusted by the Endorser and Verifier Owner before getting the Endorsements
```
dthaler: This is confusing.   There are multiple Verifiers (local and remote) and thus there can be two Endorsers, and two Verifier Owners.  So "the Endorser" and "the Verifier Owner" is ambiguous.

mcr:
Closed, replaced with #46 after rebase and edits.

PR #52, opened 2020-03-03
update the composite device picture to add the local verifier

William-PanWei:
It's more reasonable to say the local Verifier is also a Target Environment.

For the case that Lead Attester has no local Verifier, the Evidence of other Attesters is processed directly by the Attesting Environment. 
For the case that Lead Attester has a local Verifier, the Evidence of other Attesters is verified by the local Verifier (also is a Target Environment) and then claims are collected by the Attesting Environment. 

In the picture now, the arrow of 'Evidence of Attesters' points to the box of Lead Attester. Actually, for the previous case, the arrow should end to the Attesting Environment, and for the latter case, the arrow should end to the Target Environment.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
 {{composite}} depicts the conceptual data flow for a Composite Device.
 
 ~~~~
-                   .-----------------------------.
-                   |           Verifier          |
-                   '-----------------------------'
-                                   ^
-                                   |
-                                   | Composite
-                                   | Evidence
-                                   |
-.----------------------------------|-------------------------------.
-| .--------------------------------|-----.      .------------.     |
-| |                      .------------.  |      |            |     |
-| |                      |  Attesting |<--------| Attester B |-.   |
-| |                      |Environment |  |      '------------. |   |
-| |  .----------------.  |            |<----------| Attester C |-. |
-| |  |     Target     |  |            |  |        '------------' | |
-| |  | Environment(s) |  |            |<------------| ...        | |
-| |  |                |  '------------'  | Evidence '------------' |
-| |  |                |            ^     |    of                   |
-| |  |                |------------/     | Attesters               |
-| |  '----------------'  Collecting      | (via Internal Links or  |
-| |                      Claims          | Network Connections)    |
-| |                                      |                         |
-| | Lead Attester A                      |                         |
-| '--------------------------------------'                         |
-|                                                                  |
-|                       Composite Device                           |
-'------------------------------------------------------------------'
+              .-----------------------------.
+              |       Remote Verifier       |
+              '-----------------------------'
+                             ^
+                             | Evidence of
+                             | Composite Device
+                             |
+.----------------------------|----------------------------------.
+| .--------------------------|--------.                         |
+| |                   .-------------. |                         |
+| | Collecting Claims | Attesting   | |                         |
+| |         /-------->| Environment | |                         |
+| |         |         '-------------' |      .------------.     |
+| | .----------------.                |      |            |     |
+| | | Target         |                |<-----| Attester B |-.   |
+| | | Environment(s) |                |      '------------' |   |
+| | | (Optional      |                |<-------| Attester C |-. |
+| | |  Local         |                |        '------------' | |
+| | |  Verifier)     |                |<---------| ...        | |
+| | '----------------'                | Evidence '------------' |
+| |                                   | of Attesters            |
+| | Lead Attester A                   | (via Internal Links or  |
+| '-----------------------------------'  Network Connections)   |
+|                                                               |
+|                       Composite Device                        |
+'---------------------------------------------------------------'
 ~~~~
 {: #composite title="Conceptual Data Flow for a Composite Device"}
 
 In the Composite Device, each Attester generates its own Evidence by its
 Attesting Environment(s) collecting the claims from its Target Environment(s).
 The Lead Attester collects the Evidence of all other Attesters and then
-generates the Evidence of the whole Composite Attester.
+generates the Evidence of the whole Composite Device.
 
-The Lead Attester's Attesting Environment may or may not include its own
-Verifier.
-One situation is that the Attesting Environment has no local Verifier.
-In this situation, the Lead Attesting Environment simply combines the various
+Inside the entity of Attester, there may be a Target Environment
+that has the ability to verify the Evidence of other Attesters,
+it can be seen as an entity of Verifier and be called a local Verifier.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
 {{composite}} depicts the conceptual data flow for a Composite Device.
 
 ~~~~
-                   .-----------------------------.
-                   |           Verifier          |
-                   '-----------------------------'
-                                   ^
-                                   |
-                                   | Composite
-                                   | Evidence
```
William-PanWei: I didn't come up with a clear definition of 'Composite Evidence'. Besides, according to Eric's new draft, the Composite Evidence is not only applicable to the Composite Device. So I change here to 'Evidence of Composite Device'. Maybe we can create an issue about defining Composite Evidence.

draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
...
+| | | Environment(s) |                |      '------------' |   |
+| | | (Optional      |                |<-------| Attester C |-. |
+| | |  Local         |                |        '------------' | |
+| | |  Verifier)     |                |<---------| ...        | |
+| | '----------------'                | Evidence '------------' |
+| |                                   | of Attesters            |
+| | Lead Attester A                   | (via Internal Links or  |
+| '-----------------------------------'  Network Connections)   |
+|                                                               |
+|                       Composite Device                        |
```
William-PanWei: 'Composite Device' puzzles me because here limits the scenario to a single device. What about a system consisting of multiple devices? Is 'Composite System' better? A system can be a device composed of several components/modules, or a network composed of several devices, or others.

draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
...
-
-The Lead Attester's Attesting Environment may or may not include its own
-Verifier.
-One situation is that the Attesting Environment has no local Verifier.
-In this situation, the Lead Attesting Environment simply combines the various
+generates the Evidence of the whole Composite Device.
+
+Inside the entity of Attester, there may be a Target Environment
+that has the ability to verify the Evidence of other Attesters,
+it can be seen as an entity of Verifier and be called a local Verifier.
```
henkbirkholz: Maybe let me attempt a rephrase:
"The Entity that takes on the role of a Attester myay also include the role of a Verifier in its Target Environment. A Verifier residing in a Target Environment of an Attester is called a Local Verifier."

draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
...
+The claims collected from this Target Environment (i.e., the local Verifier)
+have two types. One is about itself and will finally be used to evaluate
```
henkbirkholz: The semantic context is appropriate, phrasing as 1.) "itself" and 2.) verification result is difficult to set in context.

draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
...
+the trustworthiness of the Lead Attester and itself. The other is about
+the verification results of other Attesters.
+
+One situation is that the Lead Attester has no local Verifier. In this situation,
```
henkbirkholz: Maybe start with: "in usage scenarios the Entity either includes a Lead Attester [btw I am still not sure, if we want to use this term] in one of its Target Environments or not." Then reverse and polish the following the following block of statements?

draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
...
-The other situation is that the Lead Attesting Environment has an local Verifier.
-After collecting the Evidence of other Attesters, this Attesting Environment
+The other situation is that the Lead Attester has a local Verifier.
+After collecting the Evidence of other Attesters, this local Verifier
 verifies them using Endorsements and Appraisal Policies (obtained the
 same way as any other Verifier), for evaluating these Attesters' trustworthiness.
-Then the Lead Attesting Environment combines the Attestation Results into
-the final Evidence of the whole Composite Attester which is sent off to the remote
-Verifier, which might treat the claims obtained from the local Attestation Results
+Then the Lead Attester's Attesting Environment collects the Attestation
```
henkbirkholz: "then"? A more precise conditional statement is always better :-)

draft-ietf-rats-architecture.md
William-PanWei: I'd like to double-check whether you mean "include a Lead Attester in one of its Target Environments" or "include a Local Verifier in one of its Target Environments"?

draft-ietf-rats-architecture.md
William-PanWei: It's better, I'll change the text.

draft-ietf-rats-architecture.md
```diff
@@ -296,56 +296,62 @@ each router is an Attester and the main router is the Lead Attester.
...
-generates the Evidence of the whole Composite Attester.
-
-The Lead Attester's Attesting Environment may or may not include its own
-Verifier.
-One situation is that the Attesting Environment has no local Verifier.
-In this situation, the Lead Attesting Environment simply combines the various
+generates the Evidence of the whole Composite Device.
+
+Inside the entity of Attester, there may be a Target Environment
+that has the ability to verify the Evidence of other Attesters,
```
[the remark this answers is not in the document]
William-PanWei: Yes. Your next comment is a better rephrase.

draft-ietf-rats-architecture.md
William-PanWei: I can delete this word. I have no specific condition to be stated here.

draft-ietf-rats-architecture.md
William-PanWei: Do you indicate a rephrase?

draft-ietf-rats-architecture.md
William-PanWei: The reason for using the term Lead Attester is to explicitly clarify that there is a special Attester can directly communicate with the remote Verifier while other Attesters cannot.
Maybe use "Proxy Attester"?

henkbirkholz: thank you for catching that :) In fact, what I meant to not spell wrong again in any case was:

""include a Local Verifier in one of its Target Environments"

draft-ietf-rats-architecture.md
mcr: "device" here is whatever is inside.

William-PanWei:
replaced by #58

PR #53, opened 2020-03-03
Update section 4.1 per meeting discussion

dthaler:
Provide forward references
Use "measure" in English text


dthaler:
(I had to do this right away before I forgot what I was suggesting during the meeting :)

Issue #59, opened 2020-03-05
Refine Implicit Trust concept in architecture document


gmandyam:
Current Trust Model section states:  

"a Verifier might be configured to implicitly trust firmware or even software (e.g., a hypervisor). That is, it might evaluate the trustworthiness of an application component, or operating system component or service, under the assumption that information provided about it by the lower-layer hypervisor or firmware is true. "

Would like to refine for the use of unsigned and potentially-unendorsed tokens.  If the attestation token is received over a trusted link, e.g. one that the verifier has set up with the device that is anchored in an an RoT (e.g. cellular communications link), then the verifier may implicitly trust the information sent without the evidence itself being signed.

William-PanWei:
I didn't quite get your point. Do you mean the Attester hasn't the ability to sign the Evidence, but it can be verified by the Verifier that it has an RoT?

mcr:
moved to #60.

PR #60, opened 2020-03-05
Update Trust Model with Implicit Trust Example

gmandyam:
Starting point for https://github.com/ietf-rats-wg/architecture/issues/59

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -620,7 +620,7 @@ link is anchored to a Root of Trust from an entity that established the communic
 As a result, the Verifier may be able to trust the attestation evidence from the device without
 an additional endorsement or even a cryptographically-verifiable signature of the 
 evidence.  Note that the appraisal policy employed by the Verifier may take into account the
-transient nature of the communications link security.  For instance, one the link is torn down
+transient nature of the communications link security.  For instance, once the link is torn down
 then the attestation evidence may not be trusted until it is conveyed again over a re-established link.
 
 In some scenarios, Evidence might contain sensitive information such as
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -613,8 +613,7 @@ implicitly trusted is often referred to as a Root of Trust.
 Implicit trust can also be tied to the communications link over which
 the Evidence is conveyed. As an example, the Attester and Verifier can
 communicate over a link that is established using a Root of Trust.  An example of such communications
-is over a cellular link, where the link is established by the Attester leveraging a secure
-element (e.g., SIM card).  The Verifier may have established the communications link and verified
+is over a cellular link, where the link is established by the Attester leveraging an enclave (e.g., SIM card).  The Verifier may have established the communications link and verified
 the presence of the Root-of-Trust in doing so, or may have received evidence that the communications
 link is anchored to a Root of Trust from an entity that established the communications link.
 As a result, the Verifier may be able to trust the attestation evidence from the device without
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -619,7 +619,9 @@ the preence of the Root-of-Trust in doing so, or may have received evidence that
 link is anchored to a Root of Trust from an entity that established the communications link.
 As a result, the Verifier may be able to trust the attestation evidence from the device without
 an additional endorsement or even a cryptographically-verifiable signature of the 
-evidence.
+evidence.  Note that the appraisal policy employed by the Verifier may take into account the
+transient nature of the communications link security.  For instance, one the link is torn down
+then the attestation evidence may not be trusted until it is conveyed again over a re-established link.
 
 In some scenarios, Evidence might contain sensitive information such as
 Personally Identifiable Information.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -621,7 +621,7 @@ As a result, the Verifier may be able to trust the attestation evidence from the
 an additional endorsement or even a cryptographically-verifiable signature of the 
 evidence.  Note that the appraisal policy employed by the Verifier may take into account the
 transient nature of the communications link security.  For instance, once the link is torn down
-then the attestation evidence may not be trusted until it is conveyed again over a re-established link.
+then the Evidence may not be trusted until it is conveyed again over a re-established link.
 
 In some scenarios, Evidence might contain sensitive information such as
 Personally Identifiable Information.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -612,7 +612,7 @@ implicitly trusted is often referred to as a Root of Trust.
 
 Implicit trust can also be tied to the communications link over which
 the Evidence is conveyed. As an example, the Attester and Verifier can
-communicate over a link that is established using a Root of Trust.  An example of such communicaitons
+communicate over a link that is established using a Root of Trust.  An example of such communications
 is over a cellular link, where the link is established by the Attester leveraging a secure
 element (e.g. SIM card).  The Verifier may have esablished the communications link and verified
 the preence of the Root-of-Trust in doing so, or may have received evidence that the communications
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -611,10 +611,13 @@ physically resistant to hardware tampering.  The component that is
 implicitly trusted is often referred to as a Root of Trust.
 
 Implicit trust can also be tied to the communications link over which
-the attestation evidence is conveyed. As an example, if the device and Verifier
-communicate over a link that is established using a Root of Trust (e.g. cellular
-communications link where authentication is tied to the presence of a secure element),
-then the Verifier may be able to trust the attestation evidence from the device without
+the attestation Evidence is conveyed. As an example, the Attester and Verifier can
+communicate over a link that is established using a Root of Trust.  An example of such communicaitons
+is over a cellular link, where the link is established by the Attester leveraging a secure
+element (e.g. SIM card).  The Verifier may have esablished the communications link and verified
+the preence of the Root-of-Trust in doing so, or may have received evidence that the communications
+link is anchored to a Root of Trust from an entity that established the communications link.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,14 @@ for by hardware or by ROM code, especially if such hardware is
 physically resistant to hardware tampering.  The component that is
 implicitly trusted is often referred to as a Root of Trust.
 
+Implicit trust can also be tied to the communications link over which
+the attestation evidence is conveyed. As an example, if the device and Verifier
```
William-PanWei: evidence -> Evidence
the device -> Attester

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,14 @@ for by hardware or by ROM code, especially if such hardware is
...
+communicate over a link that is established using a Root of Trust (e.g. cellular
+communications link where authentication is tied to the presence of a secure element),
```
William-PanWei: Are the Attester and Verifier different entities? If they are different, how can the Verifier know the Attester has a secure element?

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,14 @@ for by hardware or by ROM code, especially if such hardware is
...
+then the Verifier may be able to trust the attestation evidence from the device without
+an additional endorsement or even a cryptographically-verifiable signature of the 
+evidence.
```
William-PanWei: So you mean the Evidence isn't signed or encrypted?

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,14 @@ for by hardware or by ROM code, especially if such hardware is
 physically resistant to hardware tampering.  The component that is
 implicitly trusted is often referred to as a Root of Trust.
 
+Implicit trust can also be tied to the communications link over which
+the attestation evidence is conveyed. As an example, if the device and Verifier
+communicate over a link that is established using a Root of Trust (e.g. cellular
```
dthaler: nit: "e.g." should always have a comma immediately after it

draft-ietf-rats-architecture.md
dthaler: I don't think we should use the term "secure element" in lower case, as it's ambiguous whether we mean Secure Element (per GP) or something more generic

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,14 @@ for by hardware or by ROM code, especially if such hardware is
...
+communications link where authentication is tied to the presence of a secure element),
+then the Verifier may be able to trust the attestation evidence from the device without
+an additional endorsement or even a cryptographically-verifiable signature of the
```
dthaler: I don't think the word endorsement is accurate here.   Perhaps you mean signature?  (Which is already after the "or") so recommend deleting the endorsement phrase.

draft-ietf-rats-architecture.md
mcr: Dave Thaler asks for rewording to make the Implicit Trust more clear, the secure session is necessary, but not sufficient for this.  MCR asks for the example to not be parenthetical.

thomas-fossati:
There's another bit that probably deserves a small paragraph here, related to the implication of session security being "ephemeral", i.e., what happens to the evidence once the secure session goes away?

draft-ietf-rats-architecture.md
nedmsmith: Endorsements are signed (integrity protected and authenticated) by supply chain entity such as a vendor or manufacturer. Evidence is signed by an attestation key that is on the attesting device. The roles architecture distinguishes between Endorsers and Attesters for good reasons even though there can be cases where it isn't that clear that they are different. I don't think this section is trying to draw attention to one of those corner cases. Therefore, it appears the term Endorsement was used by mistake when Evidence was intended.

gmandyam: @nedmsmith 

That wasn't really the intention behind the cited text.  If the communications link is already trusted, then an endorsement may have been used in establishing the communications link.  In that case, an additional endorsement for attestation evidence is not required.

gmandyam:
Tried to address the comments in PR mod's:

a) @dthaler  - separated out example and tried to make it better-defined
b) @thomas-fossati - added 2 sentences to discuss the transience of the communications link and associated trust
c) @William-PanWei - tried to better explain how a Verifier would obtain information about how the communications link was established.  It makes me wonder if the baseline architecture would need to be extended to encompass delegated trust.
d) @nedmsmith - I may still have a misunderstanding of what an endorsement actually means in the RATS architecture.  In this example, I felt that an endorsement may have been leveraged in establishment of the communications link.  Therefore an additional attestation-specific endorsement may not be necessary.

draft-ietf-rats-architecture.md
nedmsmith: I agree with the three sentences above.

I think the text at line 616 may be what is the problem because it presumes knowledge of the secure element context is existential. The Verifier needs to be told that the root-of-trust / secure element environment exists. It can happen manually, such as an IT department physically issuing a smart card or smart device with a SE to a user, generating the key pair, update the Verifier to trust the public that was generated. Manual approaches work in IT environments where employees are subject to IT procedures. Manual processes may not scale to Internet ecosystems. Remote attestation is trying to go beyond manual processes by generating keys in devices by manufacturers and issuing machine certs that establish that a key is protected by a secure element. This seems like a basic idea that if it isn't clear might belong in the introduction?

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
 physically resistant to hardware tampering.  The component that is
 implicitly trusted is often referred to as a Root of Trust.
 
+Implicit trust can also be tied to the communications link over which
+the attestation Evidence is conveyed. As an example, the Attester and Verifier can
```
dthaler, line 614:
```suggestion
the Evidence is conveyed. As an example, the Attester and Verifier can
```

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
...
+communicate over a link that is established using a Root of Trust.  An example of such communicaitons
```
dthaler, line 615:
```suggestion
communicate over a link that is established using a Root of Trust.  An example of such communications
```

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
...
+is over a cellular link, where the link is established by the Attester leveraging a secure
+element (e.g. SIM card).  The Verifier may have esablished the communications link and verified
```
dthaler, line 617:
```suggestion
element (e.g., SIM card).  The Verifier may have established the communications link and verified
```

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
...
+the preence of the Root-of-Trust in doing so, or may have received evidence that the communications
```
dthaler, line 618:
```suggestion
the presence of the Root-of-Trust in doing so, or may have received evidence that the communications
```

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
...
+link is anchored to a Root of Trust from an entity that established the communications link.
+As a result, the Verifier may be able to trust the attestation evidence from the device without
+an additional endorsement or even a cryptographically-verifiable signature of the 
+evidence.  Note that the appraisal policy employed by the Verifier may take into account the
+transient nature of the communications link security.  For instance, one the link is torn down
```
dthaler, line 623:
```suggestion
transient nature of the communications link security.  For instance, once the link is torn down
```

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
...
+then the attestation evidence may not be trusted until it is conveyed again over a re-established link.
```
dthaler, line 624:
```suggestion
then the Evidence may not be trusted until it is conveyed again over a re-established link.
```

draft-ietf-rats-architecture.md
```diff
@@ -610,6 +610,19 @@ for by hardware or by ROM code, especially if such hardware is
 physically resistant to hardware tampering.  The component that is
 implicitly trusted is often referred to as a Root of Trust.
 
+Implicit trust can also be tied to the communications link over which
+the attestation Evidence is conveyed. As an example, the Attester and Verifier can
+communicate over a link that is established using a Root of Trust.  An example of such communicaitons
+is over a cellular link, where the link is established by the Attester leveraging a secure
+element (e.g. SIM card).  The Verifier may have esablished the communications link and verified
+the preence of the Root-of-Trust in doing so, or may have received evidence that the communications
+link is anchored to a Root of Trust from an entity that established the communications link.
```
dthaler, line 619: Need more explanation about what "anchored to a Root of Trust" means.  To trust the evidence, you have to have some assurance that the evidence is correct, and wasn't just fed to the communication link endpoint by some untrusted party.

draft-ietf-rats-architecture.md
gmandyam: The use of 'endorsement' is deliberate.  Any endorsement would be applied to assertions associated with the communications link, not the attestation evidence.

draft-ietf-rats-architecture.md
gmandyam: What about the following:

"... or may have received evidence that the communications link is anchored to a Root of Trust from an entity that established the communications link.  In the latter case, anchoring the communications link to a Root of Trust would mean that  this entity only established a communications link with the attesting device after verifying that the device had a Root of Trust."

draft-ietf-rats-architecture.md
gmandyam: Changes to "enclave" as per RFC 4949

draft-ietf-rats-architecture.md
dthaler: That wasn't obvious from "additional".  Need to reword.

mcr:
Closed in favour of PR#82.

Issue #50, opened 2020-03-03
Is there a need to have a 'basis' Topological Model?


William-PanWei:
~~~~
+-------------+            +-------------+               +-------------+
|             |----------->|             |-------------->|             |
|   Attester  |  Evidence  |   Verifier  |  Attestation  |   Relying   |
|             |            |             |     Result    |    Party    |
+-------------+            +-------------+               +-------------+
                           Compare Evidence              Compare Attestation
                           against Appraisal             Result against
                           Policy                        Appraisal Policy
~~~~
In both the passport and background-check models, the Attester communicates directly with the Relying Party to access some resources. But in some cases, the Attester doesn't need to access resources from the Relying Party.
Assume that the Attester is a network device (router, switch, etc.), and the Relying Party is the network management system (NMS) of the network operator. In such a case, the NMS monitors the health of the device by retrieving the Attestation Result of the device from a Verifier. If the device fails the attestation process, the NMS can take countermeasures like quarantining it. So the figure above depicts this situation.

William-PanWei:
Is this topological model covered by the passport model or background-check model?

mcr:
The above flow is always the information flow.  The passport model and background-check are models for how the conveyance works.

mcr:
Also, we had that diagram in the architecture documents, and we have the above diagram as figure 1 already.

William-PanWei:
> The above flow is always the information flow. The passport model and background-check are models for how the conveyance works.

This is also a way in which the conveyance can work. I add two sentences to describe this basic one without duplicating the diagram.

mcr:
We conclude that there is not a need for this additional topological flow.

PR #62, opened 2020-03-06
add two sentences to describe the basic Topological Model, address #50

William-PanWei:
describe the basic Topological Model without complicating the draft

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -425,7 +425,7 @@ in {{trustmodel}} is also suitable for the local verifier.
 
 {{dataflow}} shows a basic model for communication between an Attester,
 a Verifier, and a Relying Party. The Attester sends its Evidence to the Verifier
-for appraisal, and the Verifier sends the Attestation Results to the Relying Party.
+for appraisal, and the Relying Party gets the Attestation Results from the Verifier.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -423,8 +423,10 @@ in {{trustmodel}} is also suitable for the local verifier.
 
 # Topological Models {#overview}
 
-There are multiple possible models for communication between an Attester,
-a Verifier, and a Relying Party.  This section includes some reference models,
+{{dataflow}} shows a basic model for communication between an Attester,
+a Verifier, and a Relying Party. The Attester sends its Evidence to the Verifier
+for appraisal, and the Verifier sends the Attestation Results to the Relying Party.
```
dthaler: I disagree with the use of "sends" with Verifier as the subject.   That implies how a particular conveyance protocol works, which is too restrictive and not the point of this section in my opinion.

William-PanWei: OK. I only want to clarify that a basic communication model will be the Attester communicates with the Verifier and the Verifier communicates with the Relying Party.
So how about to say "the Relying Party gets the Attestation Results from the Verifier"?

henkbirkholz:
I am uncertain about the state of this PR. @dthaler, were your comments addressed?

PR #70, opened 2020-03-17
Reference diagram showing two types of environments

dthaler:

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -345,6 +345,9 @@ cryptographic functions, such as signing or cipher algorithms, to
 create Evidence. Examples of environments that can be used as
 Attesting Environments include Trusted Execution Environments (TEE),
 embedded Secure Elements (eSE), or Hardware Security Modules (HSM).
+An execution environment may not, by default, be capable of claims collection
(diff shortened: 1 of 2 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -346,6 +346,12 @@ create Evidence. Examples of environments that can be used as
 Attesting Environments include Trusted Execution Environments (TEE),
```
nedmsmith: Attesting environments may include...
<the reason for adding 'may' is because not all instances of environments listed are capable of securely collection claims for a given Target environment. Attesting environments (typically?) are built with the expected target environment in mind. Maybe we need more clarification in the text?>

dthaler:
I don't understand the text Ned added.

nedmsmith:
The existing text listed HSM and TEE as examples of Attesting Environments. However, an HSM that only provides crypto acceleration and key storage wasn’t designed for claims collection. Therefore, it isn’t an Attesting Environment (necessarily).

The text I added explains that the Attesting Environment is designed in such a way that it is capable of claims collection. For example, if the HSM was designed to have exposed a shared memory region that is shared with the Target Environment’s executable code pages.

henkbirkholz:
Will fix broken kramdown after pull.

PR #74, opened 2020-03-24
Define claim

mcr:


draft-ietf-rats-architecture.md
```diff
@@ -123,6 +123,10 @@ Attester:
 
 : An entity whose attributes must be appraised in order to determine whether the entity is considered trustworthy, such as when deciding whether the entity is authorized to perform some operation
 
+Claim:
+: A piece of information asserted about a subject. (Compare /claim/ in {{RFC7519}})
```
mcr: "A piece of asserted information, often in the form of a name/value pair."

mcr: Thomas Hardjono
This is what th SAML standard says: An assertion is a package of information that supplies zero or more statements made (about subject)

nedmsmith: I don't feel strongly about using 'subject' but it isn't used anywhere else whereas we use 'entity' elsewhere. If the reader knows we meant 'entity' even though we used 'subject' then it is OK. However, it might be just as easy to say entity and not introduce the question in the readers mind whether subject has special meaning or not. We use entity 29 times already. We use 'subject' once but for a different meaning.

draft-ietf-rats-architecture.md
```diff
@@ -133,6 +123,10 @@ Attester:
 
 : An entity whose attributes must be appraised in order to determine whether the entity is considered trustworthy, such as when deciding whether the entity is authorized to perform some operation
 
+Claim:
```
William-PanWei: One minor question: if we define the term Claim, do we need to further change the 'claim' now in the whole doc to capitalized 'Claim'?

draft-ietf-rats-architecture.md
```diff
@@ -133,6 +123,10 @@ Attester:
...
+: A piece of asserted information, often in the form of a name/value pair.
```
nedmsmith: The word 'asserted' is ambiguous / dissonant - leading the reader to wonder if this is a loaded word - but without context or resolution. I also don't know if by saying to 'compare' to JWT use of 'claim' whether this spec uses it synonymously or if the reader is supposed to look for and internalize how they differ.

Being explicit is better IMO. "A piece of information, often in the form of a name/value pair, that can be asserted as Evidence, Endorsements or Attestation Results."

draft-ietf-rats-architecture.md
```diff
@@ -133,6 +123,10 @@ Attester:
...
+(Compare /claim/ in {{RFC7519}})
```
nedmsmith: "Note: use of 'claim' in this document differs from RFC7519 in that this document doesn't imply a particular encoding"

draft-ietf-rats-architecture.md
```diff
@@ -85,33 +85,23 @@ remote attestation procedures (RATS).
...
-assessed via evidence created. This is a subtle difference and being
-familiar with the difference is crucial for using this document.
-Additionally, the concepts of freshness and trust relationships with
-respect to RATS are elaborated on to enable implementers in order to choose
-appropriate solutions to compose their Remote Attestation Procedures.
+Trust is a decision being made.
+Trustworthiness is a quality that is assessed via evidence created.
+This is a subtle difference and being familiar with the difference is crucial for using this document.
+Additionally, the concepts of freshness and trust relationships with respect
+to RATS are elaborated on to enable implementers in order to choose appropriate solutions to compose their Remote Attestation Procedures.
```
nedmsmith: I can't identify any changes besides formatting. Was there significance to the above highlighted differences?

draft-ietf-rats-architecture.md
```diff
@@ -85,33 +85,23 @@ remote attestation procedures (RATS).
...
-
-This documents defines a flexible architecture with corresponding roles
-and their interaction via conceptual messages. Additionally, this document
-defines a universal set of terms that can be mapped to various existing
-and emerging Remote Attestation Procedures. Common role compositions and
-data flows, such as the "Passport Model" and the "Background-Check Model"
-are illustrated to enable readers of this document to map their current
-and emerging solutions to the architecture provided and the corresponding
-terminology defined. A common terminology that provides a well-understood semantic meaning to
+In Remote Attestation Procedures (RATS), one peer (the "Attester") produces believable information about itself - Evidence - to enable a remote peer (the "Relying Party") to decide whether to consider that Attester a trustworthy peer or not.
```
dthaler, line 88: This is hard to review since the line breaks were changed to combine many lines into few lines.
Can you revert the line break changes and keep source lines to a short length like before, so
github diffs are easier?

draft-ietf-rats-architecture.md
dthaler: +1, please revert source line break changes

PR #75, opened 2020-03-26
Time considerations

dthaler:
Add time considerations into an appendix and reference it from the freshness section.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -912,8 +912,8 @@ or might be defined relative to some other timestamp or timeticks counter.
 | ID | Event                       | Explanation of event
 |----|-----------------------------|-----------------------
 | VG | Value generation            | A value to appear in a claim was created
-| VN | Nonce sent by Verifier      | A random number not predictable to an Attester is sent by a Verifier
-| VR | Nonce sent by Relying Party | The nonce is sent by a Relying Party
+| NS | Nonce sent                  | A random number not predictable to an Attester is sent
+| NR | Nonce relayed               | The nonce is relayed to an Attester by enother entity
 | EG | Evidence generation         | An Attester collects claims and generates Evidence
 | ER | Evidence relayed            | A Relying Party relays Evidence to a Verifier
 | RG | Result generation           | A Verifier appraises Evidence and generates an Attestation Result
@@ -995,7 +995,7 @@ are synchronized.
         |                                 |               |
         ~                                 ~               ~
         |                                 |               |
-        |<---Nonce1--------------------time(VN)           |
+        |<---Nonce1--------------------time(NS)           |
      time(EG)                             |               |
         |----Evidence-------------------->|               |
         |     {Nonce1, time(EG)-time(VG)} |               |
@@ -1004,7 +1004,7 @@ are synchronized.
         |     {time(RX)-time(RG)}         |               |
         ~                                                 ~
         |                                                 |
-        |<---Nonce2------------------------------------time(VR)
+        |<---Nonce2------------------------------------time(NS')
      time(RR)
         |----Attestation Result{time(RX)-time(RG)}---->time(RA)
         |    Nonce2, time(RR)-time(EG)                    |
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -897,3 +899,210 @@ Thomas Hardjono created older versions of the terminology section in collaborati
...
+# Appendix A: Time Considerations {#time-considerations}
+
+The table below defines a number of relevant events, with an ID that
+is used in subsequent diagrams.  The times of said events might be
+defined in terms of an absolute clock time such as Coordinated Universal Time, 
+or might be defined relative to some other timestamp or timeticks counter.
+
+| ID | Event                       | Explanation of event
+|----|-----------------------------|-----------------------
+| VG | Value generation            | A value to appear in a claim was created
```
nedmsmith: I think there is a difference between the time a value was generated/created and the time it was collected into a claim. Example: FW is updated on Tuesday but Attesting environment does't run until Wednesday. If the objective is to identify all the points where time could play a role, VG seems to conflate two events.

draft-ietf-rats-architecture.md
```diff
@@ -897,3 +899,210 @@ Thomas Hardjono created older versions of the terminology section in collaborati
...
+| VN | Nonce sent by Verifier      | A random number not predictable to an Attester is sent by a Verifier
+| VR | Nonce sent by Relying Party | The nonce is sent by a Relying Party
+| EG | Evidence generation         | An Attester collects claims and generates Evidence
+| ER | Evidence relayed            | A Relying Party relays Evidence to a Verifier
+| RG | Result generation           | A Verifier appraises Evidence and generates an Attestation Result
+| RR | Result relayed              | A Relying Party relays an Attestation Result to a Relying Party
+| RA | Result appraised            | The Relying Party appraises Attestation Results
+| OP | Operation performed         | The Relying Party performs some operation requested by the Attester.  For example, acting upon some message just received across a session created earlier at time(RA).
```
nedmsmith: In a hybrid use case involving multiple Relying Parties, there will be a case where the first RP may issue a result expiry (RX) that is different from the RX supplied by the Verifier. Maybe it is better name it (VX) for Verifier use and allow (RX) for use by Relying Parties?

draft-ietf-rats-architecture.md
```diff
@@ -897,3 +899,210 @@ Thomas Hardjono created older versions of the terminology section in collaborati
...
+        |                    |        {time(RX)}         |
+        ~                    ~                           ~
+        |                    |                           |
+        |                 time(OP)                       |
+~~~~
+
+The time considerations in this example are equivalent to those
+discussed under Example 1 above.
+
+## Example 4: Nonce-based Background-Check Model Example
```
nedmsmith: Shouldn't 'Nonce' be returned to Verifier from RP?

dthaler:
For readability, I have removed extraneous event times and only kept the ones used in the discussion after at least one diagram.
My goal is to keep the discussion easy to understand and not include excess details or complexity that could be left to other docs.

draft-ietf-rats-architecture.md
dthaler: My opinion: I don't think it should be a goal to identify "all the points where time could play a role", just "all the points where the discussion in this section needs to reference in some specific security check for freshness/recency".

That's why I removed events that were not referenced in any check in any example.   Any request to add another row should, in my view, come with proposed text that includes a specific security check equation that requires it.

draft-ietf-rats-architecture.md
dthaler: See earlier response about not adding rows unless/until there is text explaining what security test it's required for.  Feel free to suggest text explaining what security threat is relevant and what check some party would do to mitigate it.

draft-ietf-rats-architecture.md
dthaler: Yes thanks.  Fixed diagram.

draft-ietf-rats-architecture.md
nedmsmith: The examples use RATS Roles context to show relevance of the time consideration. The roles architecture doesn't elevate Attesting Environment and Target Environment to the same status (not arguing to do that either). There is however time considerations that apply between TE and AE interactions. The security consideration (which may be contrived) is that the target environment may exhibit one set of claims at value creation and a different set at value collection / measurement. This could happen if attack code has control of the TE and presents one set of claims at time of collection then reverts back to the other set (and presumably continues execution with the attacker's claims). I can convince myself that if the AE and TE were secure by design, that the scenario shouldn't happen. However, it is a common mistake that measurement of code is taken before it is loaded, but modified after being loaded, but the collected evidence always reports what was though to be loaded not what is actually loaded. If it is important to capture this common mistake. Maybe it belongs in security considerations?

draft-ietf-rats-architecture.md
William-PanWei: According to Ned's assumption, here is the case: Verifier -> RP1 -> RP2. The verifier gives a result expiry time in the Attestation Result to RP1. RP2 gives another result expiry to RP2. So RP2 has 2 result expiry times. Maybe Ned wants to emphasize that RP2 has to make a choice between these two times.
But RP can also have other expiry policies that come from other sources, e.g., the administrator configures an expiry time. I think it's up to the implementation to consider how to make decisions among all the policies. What is important here, however, is that the Attestation Result can contain an ‘expiry’ claim.

nedmsmith: I think you meant "RP1 gives another result expiry to RP2". I suppose it isn't clear whether RP1 can supply RX or if it only can supply RR. If RX is only a function of a Verifier then the use case doesn't make sense. If RX can be asserted by RP to other RPs, then that implies RX can be a "Remediation action" from a RP; which we said might be OK for TEEP arch to define but is likely out of scope for RATS. I prefer to place limits on "who can say what" to limit combinatorics and therefore limits the potential for insecure states. Dave T. seems to be saying that unless there is a use case, it doesn't make sense to define the freshness variable. But even with already defined variables it is possible to find insecure states unless some other constraints are acknowledged. Maybe both perspectives can be held simultaneously? 

Even if we limit RX to Verifier only. An (out-of-scope?) RP1 to RP2 flow could define different RX (RX2) that leaves RP2 in the same undecidability state that Pan Wei points out. But I don't think the arch can prevent the possibility for unsecure states. Nevertheless, I'm still an advocate for specifying which entity is expected to make which freshness claim. Put the burden of proof on the use case to justify when it makes sense for other entities to make the freshness claim.

mcr:
@henkbirkholz will look into additions around purple items. 
@nedmsmith to work on moving table to section 9 (freshness)

Issue #73, opened 2020-03-17
What are "role compositions"?


nedmsmith:
The introduction section uses the phrase "role compositions and data flows" then mentions "Passport Model" as an example of a role composition. 
In section 4.3 on Composite Device the second to last section describes "combinations of roles" using an example involving a Composite Device.

I think the architecture should have crisper understanding of two concepts a) "topological models" (as defined by the section 5 heading that describes Passport, BKC and Hybrid; b) "role composition" which ought to refer to the "combining of roles" by a given entity.

The terms "role composition" and "combining of roles" are too similar for them to have different meanings and "role composition" as an alias for "topological models" is unnecessary. 

I suggest pulling the second to the last paragraph in section 4.3 into a new section 4.4 titled "Role Composition" that has a paragraph defining role composition conceptually followed by the paragraph taken from section 4.3.

Suggested Text for section 4.4:
RATS roles are implemented by entities (e.g., devices). A simple realization of the architecture associates a single role to a single entity; for example, the topological models described in section 5.1 and 5.2 follow this convention. However, entities may host multiple roles as a way to accommodate more sophisticated use cases and deployment scenarios. An entity that both connects to a wide-area network and to a system bus may play both Attester and Verifier roles. As a system bus entity, a Verifier consumes Evidence from other devices connected to the system bus that implement an Attester role. As a wide-area network connected entity, it may implement an Attester role. The entity, as a system bus Verifier, may choose to fully isolate its role as a wide-area network Attester. Alternatively, attestation results might reflect an aggregation of system bus operational state claims in the form of Evidence. 

For example, in the Composite Device scenario, the lead Attester...
<paste remainder of paragraph from section 4.3>

Additionally, the introduction section should change "role composition" terminology to "topological models" as it references Passport.

mcr:
"Role Composition" will be worked out of document.

nedmsmith:
Also addressed in Issue55. I recommend closing this issue.

dthaler:
"role composition" no longer appears in the doc.  I agree this can be closed.

PR #77, opened 2020-03-27
addressed a better differentiation between "combining roles" and "topological models"

henkbirkholz:
* added 4.4 "Roles and Entities",
* also removed the last occurrence of the artifact "role composition", and
* retained section 4.3 as is and added the usage scenario from issue #73

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -637,7 +637,7 @@ in this document. As a result, the entity can participate as a constituent of
 the RATS architecture. Additionally, an entity can aggregate more than one
 role into itself. These collapsed roles combine the duties of multiple roles.
 In these cases, interaction between these roles do not necessarily use the
-Internet Protocol. They can, e.g., using a loopback device or other IP-based
+Internet Protocol. They can be using a loopback device or other IP-based
 communication between separate environments, but they do not have to.
 Alternative channels to convey conceptual messages include sockets, GPIO
 interfaces, local busses, or hypervisors. This conveyance is sometime
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -91,7 +91,7 @@ The Verifier appraises Evidence via Appraisal Policies and creates the Attestati
 
 This documents defines a flexible architecture with corresponding roles and their interaction via conceptual messages.
 Additionally, this document defines a universal set of terms that can be mapped to various existing and emerging Remote Attestation Procedures.
-Common role compositions and data flows, such as the "Passport Model" and the "Background-Check Model" are illustrated to enable readers of this document to map their current and emerging solutions to the architecture provided and the corresponding terminology defined.
+Common topological models and the data flows associated with them, such as the "Passport Model" and the "Background-Check Model" are illustrated. The purpose is to enable readers of this document to map their current and emerging solutions to the architecture provided and the corresponding terminology defined.
 A common terminology that provides a well-understood semantic meaning to
 the concepts, roles, and models in this document is vital to create
 semantic interoperability between solutions and across different platforms.
@@ -475,6 +475,14 @@ whether in the same conveyance protocol as the Evidence or not,
 to the outside Verifier.
 
 In this situation, the trust model described in {{trustmodel}} is also suitable for this inside Verifier.
+
+## Roles and Entities
+
+An entity in the RATS architecture includes at least on of the roles defined in this document. As a result, the entity can participate as a constituent of the RATS architecture. Additionally, an entity can aggregate more than one role into itself. These collapsed roles combine the duties of multiple roles. In these cases, interaction between these roles do not necessarily use the Internet Protocol. They can, e.g. using a loopback device or other IP-based communication between separate environments, but they do not have to. Alternative channels to convey conceptual messages include sockets, GPIO interfaces, local busses, or hypervisors. This conveyance is sometime referred to as "local conveyance". This type of conveyance is typically found in Composite Devices. Most importantly, these conveyance methods are out-of-scope of RATS, but they are presumed to exist in order to convey conceptual messages appropriately between roles.
+
+An exemplary usage scenario: An entity that both connects to a wide-area network and to a system bus taking on both the Attester and Verifier roles. As a system bus entity, a Verifier consumes Evidence from other devices connected to the system bus that implement Attester roles. As a wide-area network connected entity, it may implement an Attester role. The entity, as a system bus Verifier, may choose to fully isolate its role as a wide-area network Attester. Alternatively, attestation results might reflect an aggregation of system bus operational state claims in the form of Evidence.
+
+In essence, an entity that combines more than one role also creates and consumes the corresponding conceptual messages as defined in this document.
 
 # Topological Models {#overview}
 
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -646,10 +646,7 @@ in Composite Devices. Most importantly, these conveyance methods are
 out-of-scope of RATS, but they are presumed to exist in order to convey
 conceptual messages appropriately between roles.
 
-The following paragraph elaborates on an exemplary usage scenario:
-
-An entity that both connects to a wide-area
-network and to a system bus taking on both the Attester and Verifier roles.
+For example, an entity that both connects to a wide-area network and to a system bus is taking on both the Attester and Verifier roles.
 As a system bus entity, a Verifier consumes Evidence from other devices
 connected to the system bus that implement Attester roles. As a wide-area
 network connected entity, it may implement an Attester role. The entity, as a
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -640,8 +640,7 @@ In these cases, interaction between these roles do not necessarily use the
 Internet Protocol. They can be using a loopback device or other IP-based
 communication between separate environments, but they do not have to.
 Alternative channels to convey conceptual messages include function calls, sockets, GPIO
-interfaces, local busses, or hypervisor calls. This conveyance is sometime
-referred to as "local conveyance". This type of conveyance is typically found
+interfaces, local busses, or hypervisor calls. This type of conveyance is typically found
 in Composite Devices. Most importantly, these conveyance methods are
 out-of-scope of RATS, but they are presumed to exist in order to convey
 conceptual messages appropriately between roles.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -639,7 +639,7 @@ role into itself. These collapsed roles combine the duties of multiple roles.
 In these cases, interaction between these roles do not necessarily use the
 Internet Protocol. They can be using a loopback device or other IP-based
 communication between separate environments, but they do not have to.
-Alternative channels to convey conceptual messages include sockets, GPIO
+Alternative channels to convey conceptual messages include function calls, sockets, GPIO
 interfaces, local busses, or hypervisors. This conveyance is sometime
 referred to as "local conveyance". This type of conveyance is typically found
 in Composite Devices. Most importantly, these conveyance methods are
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -650,8 +650,7 @@ As a system bus entity, a Verifier consumes Evidence from other devices
 connected to the system bus that implement Attester roles. As a wide-area
 network connected entity, it may implement an Attester role. The entity, as a
 system bus Verifier, may choose to fully isolate its role as a wide-area
-network Attester. Alternatively, Attestation Results might reflect an
-aggregation of system bus operational state claims in the form of Evidence.
+network Attester. 
 
 In essence, an entity that combines more than one role also creates and
 consumes the corresponding conceptual messages as defined in this document.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -478,12 +478,12 @@ In this situation, the trust model described in {{trustmodel}} is also suitable 
 
 ## Roles and Entities
 
-An entity in the RATS architecture includes at least on of the roles defined
+An entity in the RATS architecture includes at least one of the roles defined
 in this document. As a result, the entity can participate as a constituent of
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -476,6 +476,14 @@ to the outside Verifier.
 
 In this situation, the trust model described in {{trustmodel}} is also suitable for this inside Verifier.
 
+## Roles and Entities
+
+An entity in the RATS architecture includes at least on of the roles defined in this document. As a result, the entity can participate as a constituent of the RATS architecture. Additionally, an entity can aggregate more than one role into itself. These collapsed roles combine the duties of multiple roles. In these cases, interaction between these roles do not necessarily use the Internet Protocol. They can, e.g. using a loopback device or other IP-based communication between separate environments, but they do not have to. Alternative channels to convey conceptual messages include sockets, GPIO interfaces, local busses, or hypervisors. This conveyance is sometime referred to as "local conveyance". This type of conveyance is typically found in Composite Devices. Most importantly, these conveyance methods are out-of-scope of RATS, but they are presumed to exist in order to convey conceptual messages appropriately between roles.
```
dthaler: Line is too long in the markdown, please split into multiple source lines to make diffs/commenting easier.

draft-ietf-rats-architecture.md
```diff
@@ -476,6 +476,14 @@ to the outside Verifier.
 
 In this situation, the trust model described in {{trustmodel}} is also suitable for this inside Verifier.
 
+## Roles and Entities
```
dthaler: My opinion: I don't like trying to make "Entities" some formal term.   I suspect "implementation" works in most cases in the text below

draft-ietf-rats-architecture.md
dthaler: typo: "on" -> "one"

dthaler: Always put a comma after e.g.

draft-ietf-rats-architecture.md
```diff
@@ -476,6 +476,14 @@ to the outside Verifier.
...
+
+An exemplary usage scenario: An entity that both connects to a wide-area network and to a system bus taking on both the Attester and Verifier roles. As a system bus entity, a Verifier consumes Evidence from other devices connected to the system bus that implement Attester roles. As a wide-area network connected entity, it may implement an Attester role. The entity, as a system bus Verifier, may choose to fully isolate its role as a wide-area network Attester. Alternatively, attestation results might reflect an aggregation of system bus operational state claims in the form of Evidence.
```
dthaler: "attestation results" -> "Attestation Results"

dthaler: "An exemplary usage scenario:" is hard to read for me since it's not a sentence or a section heading or a dictionary entry

nedmsmith: I'm experiencing cognitive dissonance trying to relate the roles to both a bus connected example and a network connected example. It seems to be focusing on the composite device but highlighting the case where composite attesters could be network connected. However, I'm not certain based on this text. But otherwise, why wouldn't that discussion belong in the section on composite device?

I think the goal is to illustrate use cases where roles can be combined. Maybe sticking to a few intuitive examples is best. e.g., verifier + relying party, endorser + verifier and maybe verifier+owner. Combining attester and verifier is problematic because the trust model objective is to separate attester from verifier.

draft-ietf-rats-architecture.md
nedmsmith: Agree that "Entity" isn't needed in the section heading. But it is helpful to separate the role from the thing that implements the role. Entity is the word most commonly used in the arch to refer to this thing. So it is OK to use it here.

draft-ietf-rats-architecture.md
henkbirkholz: TBH, I just copied your description from #73 into here. I probably set the wrong context?

henkbirkholz: I assume line 401 is okay?

nedmsmith: Do you mean line 481?
I think everything after "interaction between these roles do not necessarily use the Internet Protocol" is not about role interactions but entity interactions. I don't know if entity interactions belongs or if it does should be a separate section or at least separate paragraph. The main point of line/paragraph 481 is that an entity can host (combine) multiple roles. When that happens, the role messages between roles on peer entities remains true to the roles architecture. For example, if Verifier and RP are combined. Evidence still flows to Verifier and RP still consumes Appraisal policy for Attestation Results (even though there isn't an Attestation Results flow exposed externally to the entity).

draft-ietf-rats-architecture.md
henkbirkholz: There are various uses of entities throughout the document, including composite device. We decided against "device" in favor of "entity" and in the context of this document the use of this term is now well established. We do not want to provide implementation guidance here, but an architectural concept "on how roles can be collapsed" which is important to know to understand for the solution I-Ds because it allows for the omission of internet protocols.

I see that Dave and Ned agree, but this is one of the rare cases in which I am not convinced. I actually think a focus on implementation guidance distracts from the important parts (such as reuse of key material across roles, etc.).

draft-ietf-rats-architecture.md
henkbirkholz: fixed

henkbirkholz: fixed

draft-ietf-rats-architecture.md
henkbirkholz: After "staring" at the diff a bit, attestation results -> Attestation Results is now fixed. Did I miss something?

henkbirkholz: "The following paragraph elaborates on an exemplary usage scenario:" is now followed by a paragraph.

henkbirkholz: It remains an interaction between roles, the means are changing though. Even if all roles would be collapsed, they'd still interact, "locally". Nothing changes architecturally, I think. I consequence, this is not intended to be about entity interaction. Maybe we should simply spell that out?

draft-ietf-rats-architecture.md
dthaler: When did we "decide against device in favor of entity"?   I don't think I agree with that.

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+# Roles and Entities
+
+An entity in the RATS architecture includes at least one of the roles defined
+in this document. As a result, the entity can participate as a constituent of
+the RATS architecture. Additionally, an entity can aggregate more than one
+role into itself. These collapsed roles combine the duties of multiple roles.
+In these cases, interaction between these roles do not necessarily use the
+Internet Protocol. They can, e.g., using a loopback device or other IP-based
+communication between separate environments, but they do not have to.
+Alternative channels to convey conceptual messages include sockets, GPIO
```
dthaler, line 642: What about a simple function call within the same process?

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+interfaces, local busses, or hypervisors. This conveyance is sometime
+referred to as "local conveyance". This type of conveyance is typically found
```
dthaler, line 644: I'm not sure k we need to define a term for this.

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+in Composite Devices. Most importantly, these conveyance methods are
+out-of-scope of RATS, but they are presumed to exist in order to convey
+conceptual messages appropriately between roles.
+
+The following paragraph elaborates on an exemplary usage scenario:
```
dthaler, line 649: I still find the sentence above odd, rather than combining an intro sentence into the paragraph below.

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+
+An entity that both connects to a wide-area
+network and to a system bus taking on both the Attester and Verifier roles.
```
dthaler, line 652: Above is not a complete sentence

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+As a system bus entity, a Verifier consumes Evidence from other devices
```
dthaler, line 653: Is this referring to a Composite Device example?  If so, move it to that section, or maybe it's already covered there.

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
 {: #combination title="Example Combination"}
 
+# Roles and Entities
+
+An entity in the RATS architecture includes at least one of the roles defined
+in this document. As a result, the entity can participate as a constituent of
+the RATS architecture. Additionally, an entity can aggregate more than one
+role into itself. These collapsed roles combine the duties of multiple roles.
+In these cases, interaction between these roles do not necessarily use the
+Internet Protocol. They can, e.g., using a loopback device or other IP-based
```
mcr, line 640:
```suggestion
Internet Protocol. They can be using a loopback device or other IP-based
```

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+# Roles and Entities
+
+An entity in the RATS architecture includes at least one of the roles defined
+in this document. As a result, the entity can participate as a constituent of
+the RATS architecture. Additionally, an entity can aggregate more than one
+role into itself. These collapsed roles combine the duties of multiple roles.
+In these cases, interaction between these roles do not necessarily use the
+Internet Protocol. They can be using a loopback device or other IP-based
+communication between separate environments, but they do not have to.
+Alternative channels to convey conceptual messages include sockets, GPIO
```
mcr, line 642:
```suggestion
Alternative channels to convey conceptual messages include function calls, sockets, GPIO
```

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+
+An entity in the RATS architecture includes at least one of the roles defined
+in this document. As a result, the entity can participate as a constituent of
+the RATS architecture. Additionally, an entity can aggregate more than one
+role into itself. These collapsed roles combine the duties of multiple roles.
+In these cases, interaction between these roles do not necessarily use the
+Internet Protocol. They can be using a loopback device or other IP-based
+communication between separate environments, but they do not have to.
+Alternative channels to convey conceptual messages include function calls, sockets, GPIO
+interfaces, local busses, or hypervisors. This conveyance is sometime
```
mcr, line 643:
```suggestion
interfaces, local busses, or hypervisor calls. This conveyance is sometime
```

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+Alternative channels to convey conceptual messages include function calls, sockets, GPIO
+interfaces, local busses, or hypervisor calls. This conveyance is sometime
+referred to as "local conveyance". This type of conveyance is typically found
+in Composite Devices. Most importantly, these conveyance methods are
+out-of-scope of RATS, but they are presumed to exist in order to convey
+conceptual messages appropriately between roles.
+
+The following paragraph elaborates on an exemplary usage scenario:
+
+An entity that both connects to a wide-area
```
mcr, line 651:
```suggestion
For example, an entity that both connects to a wide-area
```

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,36 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+network and to a system bus taking on both the Attester and Verifier roles.
+As a system bus entity, a Verifier consumes Evidence from other devices
```
mcr, line 653:
```suggestion
For example, an entity that both connects to a wide-area network and to a system bus is taking on both the Attester and Verifier roles.
As a system bus entity, a Verifier consumes Evidence from other devices
```

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,33 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+An entity in the RATS architecture includes at least one of the roles defined
+in this document. As a result, the entity can participate as a constituent of
+the RATS architecture. Additionally, an entity can aggregate more than one
+role into itself. These collapsed roles combine the duties of multiple roles.
+In these cases, interaction between these roles do not necessarily use the
+Internet Protocol. They can be using a loopback device or other IP-based
+communication between separate environments, but they do not have to.
+Alternative channels to convey conceptual messages include function calls, sockets, GPIO
+interfaces, local busses, or hypervisor calls. This conveyance is sometime
+referred to as "local conveyance". This type of conveyance is typically found
```
mcr, line 644:
```suggestion
interfaces, local busses, or hypervisor calls. This type of conveyance is typically found
```

draft-ietf-rats-architecture.md
mcr: Could be, but that's not the point.

draft-ietf-rats-architecture.md
```diff
@@ -630,6 +630,32 @@ plans to support in the TEEP architecture {{-teep-arch}}.
...
+out-of-scope of RATS, but they are presumed to exist in order to convey
+conceptual messages appropriately between roles.
+
+For example, an entity that both connects to a wide-area network and to a system bus is taking on both the Attester and Verifier roles.
+As a system bus entity, a Verifier consumes Evidence from other devices
+connected to the system bus that implement Attester roles. As a wide-area
+network connected entity, it may implement an Attester role. The entity, as a
+system bus Verifier, may choose to fully isolate its role as a wide-area
+network Attester. Alternatively, Attestation Results might reflect an
+aggregation of system bus operational state claims in the form of Evidence.
```
mcr, line 654:
```suggestion
network Attester. 
```

PR #80, opened 2020-04-21
Make reference value discussion apply to both types of Appraisal Policy

dthaler:
Previously the text was specific to policy for Evidence, but the same
applies to policy for appraising Attestation Results, so generalizing
the text.

This also removes a redundant sentence where older text and newly merged text from this morning said the same thing.  And the diffs look larger than they really are, since some text was just moved around to be in a different order.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -325,25 +325,27 @@ might be obtained from an Endorser along with the Endorsements, or might be obta
 via some other mechanism such as being configured in the Verifier by an
 administrator.
 
-For example, for some claims the Verifier might check the values of claims in the
-Evidence against constraints specified in the Appraisal Policy for Evidence.  Such constraints might
-involve a comparison for equality against reference values, or a check for being in
-a range bounded by reference values, or membership in a set of reference values,
-or a check against values in other claims, or any other test.
-
-Such reference values might be specified as part of the Appraisal Policy for Evidence itself,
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -325,25 +325,27 @@ might be obtained from an Endorser along with the Endorsements, or might be obta
...
-involve a comparison for equality against reference values, or a check for being in
+The Relying Party uses Attestation Results by applying its own
+Appraisal Policy to make application-specific decisions such as authorization decisions.
+The Attestation Result Appraisal Policy might, for example, be configured in the Relying Party
+by an administrator.
+
+## Appraisal Policies
+
+An appraising party (i.e., the Verifier appraising evidence, or the Relying Party
+appraising an Attestation Result) checks the values of some claims
```
dthaler, line 336:
```suggestion
The Verifier appraising evidence, or the Relying Party
appraising an Attestation Result, checks the values of some claims
```

draft-ietf-rats-architecture.md
```diff
@@ -325,25 +325,27 @@ might be obtained from an Endorser along with the Endorsements, or might be obta
...
+against constraints specified in the Appraisal Policy.  Such constraints might
```
dthaler, line 337:
```suggestion
against constraints specified in its Appraisal Policy.  Such constraints might
```

Issue #84, opened 2020-04-28
FIXME in document


dthaler:
Section 3.1 of the draft says:

>   FIXME from Henk: Measurements at early stages of
>   Layered Attestation are NOT evidence yet.
>   This text does not cover that yet

This fixme needs to be fixed somehow.

dthaler:
There are also still statements like these in section 4.3:
> Henk: we are looking for a better term than UEFI/BIOS/Firmware

and

> [Henk: we might have to define what successful means in this example and beyond]

dthaler:
Regarding "we are looking for a better term than UEFI/BIOS/Firmware", what do folks think about "first mutable environment"?

Example in context:
OLD: The very first Attesting Environment has to ensure the integrity of the (U)EFI / BIOS / Firmware that initially boots up a composite device (e.g.,  a cell phone).
NEW: he very first Attesting Environment has to ensure the integrity of the first mutable environment that initially boots up a composite device (e.g.,  a cell phone).

nedmsmith:
We have a terminology convention for the thing being referred to as “UEFI/BIOS/FW/FMC etc.” called “Target Environment”. We should use our preferred abstract terminology then provide intuition in the form of examples

NEW2: The very first Attesting Environment has to ensure the integrity of the Target Environment (e.g. the BIOS, firmware or first mutable environment that initially boots up a composite device).

I don’t believe the example of a cell phone is helpful. Cell phones neither use BIOS (AFAIK) or commonly use the term ‘first mutable environment’.

dthaler:
> We have a terminology convention for the thing being referred to as “UEFI/BIOS/FW/FMC etc.” called “Target Environment”. 

No, those terms are not synonymous.   In layered attestation, the UEFI/BIOS/firmware is also an "Attesting Environment", and software above it (or started later than it) are also Target Environments, so we cannot use the term Target Environment to mean UEFI/BIOS/firmware in general.   See PR #85 for proposed use in context.

nedmsmith:
I agree with the text in #85 as it doesn’t use either Attesting Env or Target Env.
The issue I have with NEW is the asymmetrical use of terms “Attesting Env” and other stuff not “Target Env”. If use of Target Env is wrong then so is use of Attesting Env. The section on “Two Environments” makes these terms clear and their applicationin a Layered Device attestation also clear (in terms of cascaded measurement, attestations and executions).

Are you suggesting the resolution to #84 is to substitute with text from #85?

Thx,
[...]

PR #85, opened 2020-05-04
Address FIXME about term for UEFI/BIOS/Firmware

dthaler:
A proposal for addressing a FIXME as called out in issue #84

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -395,25 +395,22 @@ A very common example is elaborated on to illustrate Layered Attestation.
 {:layered #layered title="Layered Attester"}
 
 The very first Attesting Environment has to ensure the integrity of
-the (U)EFI / BIOS / Firmware that initially boots up a composite device (e.g.,
+the first mutable environment (e.g., BIOS or firmware) that initially
+boots up a composite device (e.g.,
 a cell phone).
-
-~~~
-    Henk: we are looking for a better term than UEFI/BIOS/Firmware
-~~~
 
 These Claims have to be measured securely.
 At this stage of the boot-cycle of a
 composite device, the Claims collected typically cannot be composed into Evidence.
 
-The very first Attesting Environment in this example can be a hardware component that is a Static Code Root of Trust.
+The very first Attesting Environment in this example can be a hardware component that is an immutable Root of Trust.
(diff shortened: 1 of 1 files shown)


dthaler:
Since people didn't like the previous version of this PR, I rewrote the layered attestation section to not need to define a term at all.  Now it just uses one specific example to illustrate the concept, rather than trying to combine multiple variations in the same text.

draft-ietf-rats-architecture.md
```diff
@@ -378,78 +372,58 @@ Attesting Environments are designed specifically with claims collection in mind.
...
+At this stage of the boot-cycle of the
+device, the Claims collected typically cannot be composed into Evidence.
+
+After the boot sequence is started, the BIOS conducts the
+most important and defining feature of layered attestation, which is that
+the successfully measured Target Environment B
+now becomes (or contains) an Attesting Environment for the next layer.
+This procedure in Layered Attestation is sometimes called "staging".
+It is important to note that the new Attesting Environment B cannot alter
+any Claims about its own Target Environment B, since those Claims were signed
```
thomas-fossati, line 407: When we say "signed" here are we implying a signing operation done by A in the cryptographic sense, i.e. using A's private key?  If so, I'm not sure this is the necessary precondition.  Measurements might just be "sealed" in the RoT at a certain boot stage for later use by the lead attester.

dthaler: Good point, updated.

draft-ietf-rats-architecture.md
```diff
@@ -378,78 +372,60 @@ Attesting Environments are designed specifically with claims collection in mind.
...
+
+After the boot sequence is started, the BIOS conducts the
+most important and defining feature of layered attestation, which is that
+the successfully measured Target Environment B
+now becomes (or contains) an Attesting Environment for the next layer.
+This procedure in Layered Attestation is sometimes called "staging".
+It is important that the new Attesting Environment B not be
+able to alter any Claims about its own Target Environment B. 
+This can be ensured having those Claims be either signed by Attesting
+Environment A or stored in a shielded and secure manner by Attesting
```
dthaler, line 409:
```suggestion
Environment A or stored in an untamperable manner by Attesting
```

PR #87, opened 2020-05-12
An update to Freshness

henkbirkholz:
Sry, for the included conflict to references. Should have checked master first....


draft-ietf-rats-architecture.md
```diff
@@ -330,14 +332,15 @@ implicitly trusted is often referred to as a Root of Trust.
...
-Result was generated using the latest Appraisal Policy for Evidence.
+Result was generated using the latest Appraisal Policy for Evidence -- and
+latest Evidence, accordingly.
 There is, however, always a race condition possible in that the
 state of the Attester, and the Appraisal Policy for Evidence, may
-change immediately after the Evidence or Attestation Result was generated.
+change immediately after the Evidence or Attestation Results were generated.
 The goal is merely to narrow the time window to something the Verifier
-(for Evidence) or Relying Party (for an Attestation Result) is willing to accept.
+(for Evidence) or Relying Party (for an Attestation Result) is willing to accept. Effectively, the burden of appraisal with respect to freshness is also a duty of the Verifier.
```
dthaler, line 343: This part doesn't make sense for Attestation Results in my opinion.

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
 used in claims as long as the signer can guarantee that they cannot have changed
 since boot.
 
+The term freshness is applied to both Evidence and Attestation Results in
+this document. Unfortunately, the term freshness is also a bit overloaded and
+in order to illustrate the application of the term in this document better,
+two characteristics that can be part of freshness are elaborated on below:
```
thomas-fossati, line 369:
```suggestion
two characteristics that can be part of freshness are elaborated on below,
```

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
...
+recentness and uniqueness:
+
+Recentness:
+
+: a quality of Evidence that is based on the time that has passed between
+creation of the evidence by an Attester (typically by "sealing" a Claims set
```
dthaler, line 375: In general this may not be correct since it may be about the time a claim value was generated rather than the time the evidence was generated.   Or perhaps maybe that's a third term and defining terms is not a good direction...

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
...
+via signing) and the reception of that Evidence by a Verifier or (relaying
```
thomas-fossati, line 376: "sealing via signing" seems over-specific to me (I commented similarly on #85).  I'd drop it and just leave "(typically by sealing a Claims set)"

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
...
+party).
+
+: Evidence can be not recent enough, if the Evidence created from collected
+Claims is too old. A nonce supplied during a Challenge/Response handshake and
```
thomas-fossati, line 380: I'm slightly confused: nonce is about uniqueness, not recentness.

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
...
+recentness and uniqueness:
+
+Recentness:
+
+: a quality of Evidence that is based on the time that has passed between
+creation of the evidence by an Attester (typically by "sealing" a Claims set
+via signing) and the reception of that Evidence by a Verifier or (relaying
+party).
+
+: Evidence can be not recent enough, if the Evidence created from collected
```
thomas-fossati, line 379: Unless I'm missing something big, this sentence looks like a truism to me?
I think this should state what defines the "too old" criterion.

henkbirkholz: While I understand your point of view.... I literally strongly disagree. It very much can be used specifically for recentness in the context of CHARRA.

draft-ietf-rats-architecture.md
henkbirkholz: pls, see below

draft-ietf-rats-architecture.md
henkbirkholz: I understand the specifics of a "Secure Channel" that would not require sealing. As this is covered by Giri's PR, I deliberately highlighted "the other"/"most common" way to inhibit tampering here. And I think that is okay, still, taking your comment into account.

draft-ietf-rats-architecture.md
henkbirkholz: pls, see the #88 

We have to "qualify" this step well. I am happy to read alternatives. It remains to be vital independent of phrasing, I think.

draft-ietf-rats-architecture.md
henkbirkholz: Correct. That is a content oversight! It's on my list, but I made an error here by omitting that.

Attestation Results have a certain "time interval of validity", which we have not talked about at all, yet? I assume?

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
...
+Claims is too old. A nonce supplied during a Challenge/Response handshake and
+a time-stamp based proof-of-recentness between time-synchronized peers are
+both usage scenarios in which recentness characteristics are provided to the
+Verifier.
+
+Uniqueness:
+
+: a quality of Evidence that is based on state of the Verifier that is
```
thomas-fossati, line 387: I'm not sure why this should be Evidence specific?  (This also applies to Recentness above.)

draft-ietf-rats-architecture.md
thomas-fossati: The thing I wanted to say is that any concept of "recentness" must come with a (explicit or implicit) "max-age" label, otherwise it isn't actionable.

draft-ietf-rats-architecture.md
thomas-fossati: Sorry, I can't see anything below, even after refreshing the page.

draft-ietf-rats-architecture.md
henkbirkholz: Example: a nonce provided by a Verifier in a network protocol to retrieve Evidence from an Attester that is subsequently used in creating the Evidence enables a Verifier to assert that the reply from the Attester including the evidence was created upon the just emitted request of the Verifier; therefore establishing the most common method of providing a proof-of-recentness via a challenge/response remote attestation (CHARRA).

Maybe I am missing something obvious here, but how is this mechanic not addressing the appraisal of "recent creation of evidence"?

draft-ietf-rats-architecture.md
henkbirkholz: Sry. "Below" was meant to refer to a different comment of yours on lines 376-379 (I am not sure how to ref that comment here).

draft-ietf-rats-architecture.md
henkbirkholz: The entire text I wrote was Evidence-specific. Your comment relates to Dave's on lines 340+343 ("above"). I am hesitant to extend this beyond the scope of Evidence and Attestation Results (the latter is missing). Would the scope of Evidence + Attestation Results be okay?

draft-ietf-rats-architecture.md
thomas-fossati: It's the _uniqueness_ property of the nonce that guarantees that the Evidence is fresh.  A nonce bears no recentness.

thomas-fossati:
General comment / suggestion.  I don't think we should go in great details here.  It's good to provide a bit of context and clarifying information around freshness, but I don't see the need for a big taxonomical effort :-)

I think the concepts could be condensed in a couple of paragraphs, like:
```
Freshness has two dimensions: recentness and uniqueness.

Uniqueness is conveyed via a non-repeatable value sourced, for example,
from a high quality randomness pool, a “granular enough” clock, a serial
number that doesn’t wrap-around, etc.  In the context of Evidence and
Attestation Results, uniqueness can be encoded into a "nonce" attribute.

Recentness is conveyed via either a relative or absolute timestamp - 
associated with the time of collection of claims by the attester, or the
time the attestation report is produced — and an either explicit
or implicit “expiry” value that determines the threshold between
acceptable (i.e. recent enough) and non-acceptable (i.e., too old)
values of the timestamp. It assumes time synchronisation between
producers and consumers.    In the context of Evidence and Attestation
Results, recentness can be encoded using a "timestamp" and "expiry"
attributes.
```
or similar.  Cheers!

draft-ietf-rats-architecture.md
henkbirkholz: Then no value, including the timestamp of a certificate, bears no inherent recentness.

But that is not correct in the context of the Challenge/Response handshake. Again, maybe I am missing something very obvious. Could you provide me with a protocol-based exemplary use of a nonce, generated by the challenger, where the reply that uses the nonce (most likely in a "sealed" way, such as digital signing) does not bear a proof-of-recentness by using that nonce in its reply?

henkbirkholz:
I agree that is not necessary to start a taxonomy here ). It would be very tiny in any way. I think that it would suffice to say that recentness and uniqueness are the two vital parts of freshness with respect to evidence (attestation results work differently, they are literally more like actual tokens and could come with an expiration date aka a period of validity).

draft-ietf-rats-architecture.md
henkbirkholz: An additional clarifying comment:

If you add a timestamp that was generated on a responder in evidence, the "meaning of recentness it bears" is based on a trust relationship created by an endorser that vouches for the fact that "timestamps from that attestation Evidence provenance" are to be trusted. Alternatively, a challenger can simply establish a similar bi-lateral trust relationship by conducting a Challenge/Response handshake here. It is a smaller scoped trust relationship, it is temporary, and the challenger must at least be RESTful in order not to forget that it created the nonce in the Challenge. But then - in this context - the nonce created by the challenger "bears recentness" from the point of view of the Challenger, I would think.

henkbirkholz: The max-age in this case is not included in the evidence as a value, it is simply state (like... a Verifier measuring a time interval to the best of its capability) maintained by the challenger. In that way, like a secure channel enabling you to omit a signature, the nonce enables you to omit a max-age while retaining being actionable.

thomas-fossati: > Then no value, including the timestamp of a certificate, bears inherent recentness characteristics, yes?

Not by itself.  You also need a synchronised clock and an expiry threshold to determine “freshness by recentness”, at least in the sense I intend it.

> But that is not correct in the context of the Challenge/Response handshake. Again, maybe I am missing something very obvious. Could you provide me with a protocol-based exemplary use of a nonce, generated by the challenger, where the reply that uses the nonce (most likely in a "sealed" way, such as digital signing) does not bear a proof-of-recentness by using that nonce in its reply?

This is IMO “freshness by uniqueness”, in fact it doesn’t require a clock nor an expiry to work, just a non-repeating source of byte strings.

thomas-fossati: I guess my main point here is the following: if we need to go into the semantic finesse of distinguishing between the recentness and uniqueness aspects of freshness, then we need to make it worth the words we spend by _clearly_ separating the two.

Since I've diluted it across a few different comments, it's maybe worth summing up the way I see the distinction working in practice :
- uniqueness needs a source of non-repeatable byte strings and is completely driven by the requestor;
- recentness needs synchronised clocks between requestor and responder and is conveyed via a timestamp plus an expiration threshold associated with the provided evidence / attestation result.

The problem I have with the current text when it says that nonces convey recentness is it seems to be blurring that distinction, therefore voiding the overall "semantic finesse" plan :-)

henkbirkholz: > > Then no value, including the timestamp of a certificate, bears inherent recentness characteristics, yes?
> 
> Not by itself. You also need a synchronised clock and an expiry threshold to determine “freshness by recentness”, at least in the sense I intend it.

Then I agree that without context, values never bear a meaning of recentness.



> > Then no value, including the timestamp of a certificate, bears inherent recentness characteristics, yes?
> 
> Not by itself. You also need a synchronised clock and an expiry threshold to determine “freshness by recentness”, at least in the sense I intend it.
> 
> > But that is not correct in the context of the Challenge/Response handshake. Again, maybe I am missing something very obvious. Could you provide me with a protocol-based exemplary use of a nonce, generated by the challenger, where the reply that uses the nonce (most likely in a "sealed" way, such as digital signing) does not bear a proof-of-recentness by using that nonce in its reply?
> 
> This is IMO “freshness by uniqueness”, in fact it doesn’t require a clock nor an expiry to work, just a non-repeating source of byte strings.

Well, the Verifier has to retain state, "remembering the nonce" until a certain time-out, which I think most certainly requires a clock. Otherwise, networks protocols would have a hard time.

The curious mixture of "freshness by uniqueness" is exactly the reason why I wrote the text. Typically "freshness" is used without context here. And we decided to provide more context because it is not "freshness by uniqueness" but "recentness & uniqueness" called freshness for simplicity, I think.

henkbirkholz: > I guess my main point here is the following: if we need to go into the semantic finesse of distinguishing between the recentness and uniqueness aspects of freshness, then we need to make it worth the words we spend by _clearly_ separating the two.

Absolutely agreed.

> * uniqueness needs a source of non-repeatable byte strings and is completely driven by the requestor;
> 
> * recentness needs synchronised clocks between requestor and responder and is conveyed via a timestamp plus an expiration threshold associated with the provided evidence / attestation result.

I am not entirely sure that uniqueness is always completely driven by the challenger/requester, but it most certainly can and is.

I agree that recentness is based on synchronization and timestamps in TUDA, of course. I still do not understand how that is required for a Challenge/Response handshake, because there are conveyance mechanisms today that only need a clock on the Challenger side, measuring the time between a Challenge including a nonce and a Response including a digest of the nonce (in a secure channel or digitally signed).

 

> The problem I have with the current text when it says that nonces convey recentness is it seems to be blurring that distinction, therefore voiding the overall "semantic finesse" plan :-)

You are correct that a nonce value by itself does not convey recentness :-) We wildly agree that even timestamp values do not do that. I think that is typically way harder to agree on. A nonce in a Challenge/Response interaction does convey recentness, but due to the context of the protocol mechanism - not by the value itself. I tried to always provide the context of the interaction model in the text, but i seems I was not successful with that.

thomas-fossati: > Well, the Verifier has to retain state, "remembering the nonce" until a certain time-out, which I think most certainly requires a clock. Otherwise, networks protocols would have a hard time.

I disagree, the existence of a garbage collector that sweeps allocated nonces at regular intervals (i.e., the "remembering until time-out") is just an implementation detail, it's not fundamental.

What a nonce gives you is the assurance that *when* the transaction completes -- maybe one ms after the end of the universe! ;-) -- the response you get back is uniquely bound to your original request and there is no other transaction that could have produced the same response.  This is what I mean by freshness by uniqueness.

henkbirkholz: Okay :-) Let me rephrase that. A Challenger that intends to retrieve Evidence in a finite, has a clock and that intends to obtain a notion of recentness can track a time time interval only using its clock to do this. The indicator for the duration here (duration being a time interval where the left time is an epoch starting with 0 at which the Challenge including a nonce is emitted and at the right time evidence including a digest of a nonce is received) is the time that has passed between Challenge and Response.

Very small RESTful interactions as enabled via CoAP enable this (even with using an observe option) notion of recentness - intrinsically. Retaining state for an operation is not an implementation detail, I think. It is, for example required to associate CoAP tokens. Typically protocols use time-outs to avoid waiting until the end of the universe - and I think that is not an implementation detail, too, but an interaction model capability. The binding to what you call an transaction here, which typically is a operation I think, is what is the essential part to get a notion of recentness. 

In contrast to your point, my point is that a re-use of a nonce that is cryptographically (and in this case uniquely) "bound" to a Response does not require synchronized clocks due to the "recentness" based on the uniqueness that is provided by the "arbitrary number used only once in a cryptographic communication" (nonce) intrinsically.

In consequence, I think:

1.) "recentness by uniqueness" is an over-simplification (you could receive two responses based on the same nonce, violating the concept of how to use a nonce, and still only process the most recent response via sequence or clock use), and
2.) "recentness needs synchronized clocks between requestor and responder and is conveyed via a timestamp plus an expiration threshold associated with the provided evidence / attestation result." is not correct, due to being too specific (you do not always need synchronized clocks) and unnecessary limiting (you do not necessarily need time-stamps and or expiration dates).

If I would only look at the interaction model of TUDA, I would agree! :-) But looking at charra, tuda and subscribed evidence, I think I still cannot agree.

thomas-fossati: OK, it seems like we agree to disagree then :-)

henkbirkholz: I assume there is common ground somewhere :-) Illustrating the different aspects about freshness for conceptual messages generally, might help us to converge maybe? Appraisal Policies enable a period or maybe even duration of validity for Attestation Evidence. Then synchronized clocks are a vital part here. 

Also, there might be other opinions out there?

thomas-fossati: > I assume there is common ground somewhere :-) 

As much as I'd like to, I think our paths diverge a bit too early to try and reconcile.

Re-reading your last batch of replies, though, I've found something I think you are 100% correct:

> you do not always need synchronized clocks

true!  You only need that when you the requestor depends on an _explicit_ timestamp set by the responder.

draft-ietf-rats-architecture.md
dthaler: I also think the time of reception is not particularly interesting, and the phrasing in 376 is not correct.  What's interesting is the time of appraisal, which is time(RG).

draft-ietf-rats-architecture.md
dthaler: @thomas-fossati the nonce is about freshness.  See discussion of examples 2 and 4 in sections 15.2 and 15.4 in the editors copy.

draft-ietf-rats-architecture.md
```diff
@@ -360,6 +363,42 @@ are signed.  For example, values might have been generated at boot, and then
...
+via signing) and the reception of that Evidence by a Verifier or (relaying
+party).
+
+: Evidence can be not recent enough, if the Evidence created from collected
+Claims is too old. A nonce supplied during a Challenge/Response handshake and
+a time-stamp based proof-of-recentness between time-synchronized peers are
+both usage scenarios in which recentness characteristics are provided to the
+Verifier.
+
+Uniqueness:
```
dthaler, line 385: I don't understand why this discussion is needed/relevant.   (I might understand it in the context of a specific protocol/mechanism, but I don't understand why it belongs in the architecture.)

draft-ietf-rats-architecture.md
thomas-fossati: @dthaler I certainly agree nonce is about freshness.  What I am arguing here (and in the following, long-winded discussion) is that a nonce creates freshness by "uniqueness" rather than "recentness".

draft-ietf-rats-architecture.md
henkbirkholz: As various individuals seem to have 2+ points of view on this, I'd assume this will create a severe caveat, if not illustrated and unambiguously resolved in the architecture I-D . In the end, if we do not agree on this intuitively, how is the architecture doc not in charge of preventing this - obviously now identified - cause of confusion to other readers?

draft-ietf-rats-architecture.md
```diff
@@ -330,14 +332,15 @@ implicitly trusted is often referred to as a Root of Trust.
 It is important to prevent replay attacks where an attacker replays
 old Evidence or an old Attestation Result that is no longer correct.
 To do so, some mechanism of ensuring that the Evidence and Attestation
-Result are fresh, meaning that there is some degree of assurance
+Results are fresh, meaning that there is some degree of assurance
 that they still reflect the latest state of the Attester, and that any Attestation
-Result was generated using the latest Appraisal Policy for Evidence.
+Result was generated using the latest Appraisal Policy for Evidence -- and
+latest Evidence, accordingly.
```
William-PanWei, line 338: why not just say "using the latest Appraisal Policy for the latest Evidence"

henkbirkholz:
Reading the whole exchange again - how about:

```
Freshness of the content of Conceptual Messages has two dimensions:
recentness and uniqueness.
 
Uniqueness of Conceptual Messages can be conveyed via an included,
non-repeatable value sourced, for example, from a high quality randomness pool,
a “granular enough” clock, a serial number that doesn’t wrap-around, etc.  In the
context of Evidence and Attestation Results, uniqueness can be encoded into a
"nonce" attribute.
 
Recentness of Evidence can be conveyed either via a protocol processing a
non-repeatable value (as illustrated above) in a hand-shake interaction or via a
payload including a relative or absolute timestamp associated with the time of
collection of Claims by the Attester, or the time the attestation Evidence is produced.
Recentness of Conceptual Messages can also be determined by either an explicit
or implicit “expiry” value that determines the threshold between acceptable
(i.e. recent enough) and non-acceptable (i.e., too old) values of the timestamp.
It assumes time synchronization between producers and consumers.
```

The sentences are still a bit long, but for the moment the more important question is,
if @thomas-fossati would agree with this text that I based on his proposal?

dthaler:
I don't understand why we care about uniqueness in this document.  The text does not justify why uniqueness matters for any security purpose.   Maybe a protocol can discuss it but I don't yet see why it should be mentioned in this doc.

thomas-fossati:
> The sentences are still a bit long, but for the moment the more important question is,
> if @thomas-fossati would agree with this text that I based on his proposal?

Thanks @henkbirkholz for trying to draft a compromise.

There are some things in your write up that needs further clarification and others that seem a bit redundant and can be easily trimmed.  

Among the first set I think one key clarification is the one around:
```
Recentness of Evidence can be conveyed [...] via a protocol processing a
non-repeatable value (as illustrated above) in a hand-shake interaction
```
which, IMO, falls a bit too short defining exactly how uniqueness can help with recentness, which seems to be your main point.

The way I understand it, is that nonce's non-repeatability used in a transaction involving an initiator and a responder (possibly what you call "hand-shake interaction" above?) allows the initiator to lock a specific portion of time, i.e., to define two very specific events: a begin and an end.  If the initiator happens to also have:
1) a clock, and
2) a notion of expiry

then some form of _rough_ recentness, i.e., one that:

1) can't distinguish between creation time and collection time of claims;
2) applies to the whole claims set as opposed to individual claims

can be built even _without_ a synchronised clock at the responder.

Is this what the quoted sentence really means to you?

If so, I think we can craft some text together that is short and to the point.

That said, and again thanks very much for this further edit pass, I'd be still OK with just dropping this PR altogether.

henkbirkholz:
@thomas-fossati, yes that is a way more detailed description of what my (rather short) was intended to capture :-)

draft-ietf-rats-architecture.md
```diff
@@ -330,14 +332,15 @@ implicitly trusted is often referred to as a Root of Trust.
 It is important to prevent replay attacks where an attacker replays
 old Evidence or an old Attestation Result that is no longer correct.
 To do so, some mechanism of ensuring that the Evidence and Attestation
-Result are fresh, meaning that there is some degree of assurance
+Results are fresh, meaning that there is some degree of assurance
```
nedmsmith, line 335: Intuitively freshness is the time since the request was made and the response was received. Do we need more than this?

draft-ietf-rats-architecture.md
nedmsmith: I think the uniqueness property is addressed by the credential used to authenticate the Attester to the Verifier (and subsequently used to integrity protect the Evidence messages). If the credential uniquely identifies the Attester then the Evidence is unique to the session to which the authentication applies. 
If the credential is a group credential (e.g. EPID / DAA) then the uniqueness is to the members of the group. We have other sections on Privacy and authentication of the endpoint. (Or maybe I'm wrong about the latter). But it might make sense to have a section on "Authentication of the Endpoint" as the way to talk about security properties of uniqueness / non-uniqueness?

thomas-fossati:
Freshness is dealt with in #91

PR #88, opened 2020-05-12
Yet another update to Time Considerations

henkbirkholz:
only the most essential items:
* added three more timestamp types based on TUDA, trying to generalize it a bit
* added a few periods to text that seems to express sentences


draft-ietf-rats-architecture.md
```diff
@@ -957,18 +957,22 @@ or might be defined relative to some other timestamp or timeticks counter.
...
+| RG | Result generation           | A Verifier appraises Evidence and generates an Attestation Result.
+| RP | Result push                 | A set of one or more Evidence bundles is conveyed to a Verifier with or without solicitation initially triggered by AA, in a periodic interval, or ad-hoc.
+| RR | Result relayed              | A Relying Party relays an Attestation Result to a Relying Party.
+| RA | Result appraised            | The Relying Party appraises Attestation Results.
 | OP | Operation performed         | The Relying Party performs some operation requested by the Attester.  For example, acting upon some message just received across a session created earlier at time(RA).
-| RX | Result expiry               | An Attestation Result should no longer be accepted, according to the Verifier that generated it
+| RX | Result expiry               | An Attestation Result should no longer be accepted, according to the Verifier that generated it.
 
-We now walk through a number of hypothetical examples of how
+Using the table above, the examples following below illustrate a number of hypothetical examples of how
```
thomas-fossati, line 975: Duplicate "examples".

(I really prefer how it read before.)

henkbirkholz: Yes. Redundant "example" is redundant. Alas, we cannot start standards text with "We now walk". The alternate text was written too hastily be me.

draft-ietf-rats-architecture.md
```diff
@@ -957,18 +957,22 @@ or might be defined relative to some other timestamp or timeticks counter.
...
-| VG | Value generation            | A value to appear in a claim was created
-| NS | Nonce sent                  | A random number not predictable to an Attester is sent
-| NR | Nonce relayed               | The nonce is relayed to an Attester by enother entity
-| EG | Evidence generation         | An Attester collects claims and generates Evidence
-| ER | Evidence relayed            | A Relying Party relays Evidence to a Verifier
-| RG | Result generation           | A Verifier appraises Evidence and generates an Attestation Result
-| RR | Result relayed              | A Relying Party relays an Attestation Result to a Relying Party
-| RA | Result appraised            | The Relying Party appraises Attestation Results
+| VG | Value generation            | A value to appear in a Claim was created.
+| AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
```
dthaler, line 961: This ID is never used in later sections, so should not be added unless it is used in the doc.

henkbirkholz: Which ID do you mean here, exactly? It is had to read from the diff illustrated, I think.

draft-ietf-rats-architecture.md
William-PanWei: Maybe we can say:
Using the table above, a number of hypothetical examples of how a solution might be built are illustrated below.

draft-ietf-rats-architecture.md
William-PanWei: I think Dave's meaning is that the newly added 3 events (AA, CC, and HD) are not used in later sections.

henkbirkholz: ah okay. Thx Wei. I created  a commit (https://github.com/ietf-rats-wg/architecture/commit/620ef97792096ab82df628f028b5c58c0ddbb4bc) based on that identified gap.

draft-ietf-rats-architecture.md
henkbirkholz: Fine with me. Thx! Objections or request for polish?

mcr, line 975:
```suggestion
Using the table above, a number of hypothetical examples of how a solution might be built are illustrated below.
```

draft-ietf-rats-architecture.md
```diff
@@ -1110,8 +1116,15 @@ clocks between the Attester, Verifier, and Relying Party.
...
+
+The goal is to create up-to-date and recent Evidence as soon as possible, in general. 
+
+At time(AA) the Attesting Environment is able to trigger an event (e.g. based on an Event-Condition-Action model) to create attestation Evidence as recent as possible. As soon as the Attesting Environment is aware, the new values are collected for composition into semantically enriched Claims at time Time(CC). Immediately after that, Evidence based on relevant "old" Claims and the most recent collected "new" Claims is generated at time(EG).
+
+In order to create attestation on demand due to the event trigger time(AA) at the time(EG), the Attester has be in synchronized state with respect to the Verifiers perception of global time. In order to accomplish his, the Attester has to have an identifier for time-bound recentness available at any time(HD) before time(EG) where the time interval between time(HD) and time(EG) is not too large due to time drift or other synchronization disablers (delta(time(HD),time(EG)<threshold)).
```
dthaler, line 1125: time(AA) seems to be irrelevant for any security purpose, only time(EG) is relevant, so I don't think time(AA) belongs in this doc.

time(HD) also seems very confusing.   since time(HD) and time(EG) are on the same entity, and that same entity compares against a threshold, what security threat is this trying to address?  I don't follow.

draft-ietf-rats-architecture.md
```diff
@@ -1110,8 +1116,15 @@ clocks between the Attester, Verifier, and Relying Party.
         |                 time(OP)                       |
 ~~~~
 
-The time considerations in this example are equivalent to those
-discussed under Example 1 above.
+In comparison with example 1, the time considerations in this example go into more detail with respect to the life-cycle of Claims and the requirements on time synchronization and corresponding timestamps.
+
+The goal is to create up-to-date and recent Evidence as soon as possible, in general. 
+
+At time(AA) the Attesting Environment is able to trigger an event (e.g. based on an Event-Condition-Action model) to create attestation Evidence as recent as possible. As soon as the Attesting Environment is aware, the new values are collected for composition into semantically enriched Claims at time Time(CC). Immediately after that, Evidence based on relevant "old" Claims and the most recent collected "new" Claims is generated at time(EG).
```
dthaler, line 1123: there is no time(CC) in the diagram, but I think time(CC) is irrelevant for any security purpose, only time(EG) matters and time(CC) can be ignored

draft-ietf-rats-architecture.md
henkbirkholz: In order to always create the most recent Evdence possible, the Attester has to be aware of a changed Value as soon as possible. Otherwise Evidence is outdated and could even convey a misleading picture about Attester leading to incorrect risk assessmet. Hence time(AA) is quite relevant, I think. 

There are scenarios where this is less relevant (e.g. boot-time integrity) and some scenarios where this is more relevant (e.g. run-time integrity).

I agree that time(HD) is "squeezed" into the diagram it is actually the reception of the distributed handles. I did not want to add a "distributor of the handles" the the actors in the diagram. Maybe we can address this by rewording time(HD) correspondingly? In essence, I understand why this is difficult to follow at the moment and we have to fix this.

draft-ietf-rats-architecture.md
henkbirkholz: I think that we can actually skip time(CC). It seemed important to various people, but I can see how you can work around that just with time(AA) and time(EG). Other thoughts?

draft-ietf-rats-architecture.md
mcr: Agreed that "HD" is not a good term, but a better term will be worked on.

PR #91, opened 2020-06-01
Freshness section slight reorg

thomas-fossati:
This commit slightly changes the Freshness section to reflect the
outcome of the discussion around PR #87.  (Hopefully I've managed to
capture the gist of Henk's thinking without disrupting too much the
existing flow.)

TL;DR

The key change here is the removal of reply protection as a goal in
itself.  The section now states clearly that the overall goal is instead
the identification of the point in time when attestation claims have
been signed and/or generated so that a freshness policy can be applied.
In this perspective, the fundamental property provided by a nonce-based
scheme with regards to freshness is its ability to lock the requester's
clock to the underlying protocol transaction, thus indirectly
establishing the epoch of the attestation / verification event.  By
making the nonce unique and non-predictable, the timeline locking cannot
be subverted, which provides anti-replay.  However, that is a nice
side-effect rather than the primary goal.

This commit also adds some precision around the nonce and clock
properties - unpredictability and trustworthiness respectively.


henkbirkholz:
Added consolidated text based on Thomas initial version and review from MRC

draft-ietf-rats-architecture.md
```diff
@@ -131,7 +131,7 @@ Attestation Result:
 
 Attester:
 
-: An entity (typically a device), whose Evidence must be appraised in order to infer the extent to which the Attester is considered trustworthy, such as when deciding whether it is authorized to perform some operation
+: An entity whose attributes must be appraised in order to determine whether the entity is considered trustworthy, such as when deciding whether the entity is authorized to perform some operation
```
dthaler, line 134: I believe "Evidence" is slightly better than "attributes" (but I can live with it either way) since only the attributes that appear in Evidence are actually appraised.

On "whether" vs "the extent to which", I believe the latter is more correct since the former implies it's always exactly boolean, and the latter is more liberal in terms of how may different possibilities there are.

draft-ietf-rats-architecture.md
```diff
@@ -143,28 +143,27 @@ Endorsement:
...
+: An entity that creates Endorsements that can be used to help to appraise the trustworthiness of Attesters
 
 Evidence:
 
 : A set of information about an Attester that is to be appraised by a Verifier
 
 Relying Party:
 
-: An entity, that depends on the validity of information about an Attester, for purposes of reliably applying application specific actions.  Compare /relying party/ in {{RFC4949}}
+: An entity that depends on the validity of information about another entity, typically for purposes of authorization.  Compare /relying party/ in {{RFC4949}}
```
dthaler, line 154: FYI I am ok either way, but I think "another entity" was previously changed to "an Attester" so not sure why reverting is being proposed in this PR but again I can live with it either way.

draft-ietf-rats-architecture.md
```diff
@@ -143,28 +143,27 @@ Endorsement:
...
 
 Relying Party Owner:
 
-: An entity (typically an administrator), that is authorized to configure Appraisal Policy for Attestation Results in a Relying Party
+: An entity, such as an administrator, that is authorized to configure Appraisal Policy for Attestation Results in a Relying Party.
 
 Verifier:
 
 : An entity that appraises the validity of Evidence about an Attester
- and produces Attestation Results to be used by a Relying Party
```
dthaler, line 163: Why is this line removed?

draft-ietf-rats-architecture.md
```diff
@@ -194,6 +193,12 @@ provides device identity and protected storage for measurements.
 These components perform a series of measurements, and express this with Evidence as to the
 hardware and firmware/software that is running.
 
+~~~
+    FIXME from Henk: Measurements at early stages of
+    Layered Attestation are NOT evidence yet.
+    This text does not cover that yet
+~~~
```
dthaler, line 200: We should not be adding FIXME comments into the doc at this point

draft-ietf-rats-architecture.md
```diff
@@ -373,60 +378,78 @@ Attesting Environments are designed specifically with claims collection in mind.
...
-The fact that an Attester role is composed of environments that
+The fact that an Attester role is composed of several types of environments that
 can be nested or staged adds complexity to the architectural layout of how an
-Attester can be composed and therefore has to conduct the Claims collection
-in order to create believable attestation Evidence.
-
-{{layered}} depicts an example of a device that includes (A) a BIOS stored
-in read-only memory in this example, (B) an updatable bootloader, and (C)
-an operating system kernel.
+Attester -- in itself -- is composed and therefore has to conduct the Claims collection
```
dthaler, line 386: I find the wording of "--in itself-- is composed" to be harder to read than the previous text

draft-ietf-rats-architecture.md
```diff
@@ -373,60 +378,78 @@ Attesting Environments are designed specifically with claims collection in mind.
...
-In general, the number of layers may vary by device or implementation,
-and an Attesting Environment might even have multiple Target Environments
-that it measures, rather than only one as shown in {{layered}}.
+The very first Attesting Environment has to ensure the integrity of
+the (U)EFI / BIOS / Firmware that initially boots up a composite device (e.g.,
+a cell phone).
+
+~~~
+    Henk: we are looking for a better term than UEFI/BIOS/Firmware
+~~~
```
dthaler, line 403: This seems to revert a bunch of changes we agreed to.   Maybe it needs a rebase or something?

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
 
 # Freshness {#freshness}
 
-It is important to prevent replay attacks where an attacker replays
-old Evidence or an old Attestation Result that is no longer correct.
-To do so, some mechanism of ensuring that the Evidence and Attestation
-Result are fresh, meaning that there is some degree of assurance
-that they still reflect the latest state of the Attester, and that any Attestation
-Result was generated using the latest Appraisal Policy for Evidence.
+It is important to prevent old Evidence or old Attestation Results that are no longer correct to be assessed as valid. Replay attacks are an important reason to do that. More importantly, the validity of Conceptual Messages, such as Evidence or Attestation Results, are the basis for trustworthy remote attestation procedures, in general.
```
dthaler, line 838: Please break lines into shorter source lines, like in the old text, to make diffs easier to review.

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+and reuse of both Evidence and Attestation Results, which is especially
+valuable in cases where their computation uses a substantial part of a
+resource budget (e.g., an energy budget in constrained devices).
+The goal is to narrow the time window to something the Verifier
 (for Evidence) or Relying Party (for an Attestation Result) is willing to accept.
 
-There are two common approaches to providing some assurance of freshness.
-The first approach is that a nonce is generated by a remote entity
+There are four common approaches to providing some assurance of freshness, effectively determining the epoch of the
+validity-span of, e.g., Evidence or Attestation Results.
```
dthaler, line 854: "epoch of the validity-span" is hard to parse and understand.  Can this be rephrased in layman's terms?

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
 the signer's time synchronization mechanism in order to provide such assurance.
-
-In either approach, it is important to note that the actual values in claims
-might have been generated long before the claims are signed.  If so, it is the
-signer's responsibility to ensure that the values are still correct when they
-are signed.  For example, values might have been generated at boot, and then
-used in claims as long as the signer can guarantee that they cannot have changed
+This allows associating a "rough" epoch to the Evidence or Attestation Result. 
+In this case, the epoch is said to be "rough" because 
+    a) it is always bound to the set of Claims as a whole, as opposed to having Claim being separately annotated, and
```
dthaler, line 880: "having Claim" is bad grammar.
"having Claims"?  "having a Claim"? "having each Claim"?

But which approach is the one where each Claim can be separately annotated with a timestamp?   In the original text with only two approaches, it would be part of this one.

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+In this case, the epoch is said to be "rough" because 
+    a) it is always bound to the set of Claims as a whole, as opposed to having Claim being separately annotated, and 
+    b) it is not possible to distinguish between time of collection from the time of creation of the Claims set.
+
+## Interval Timekeeping using Nonces
+
+A third approach is a variant of the first; the difference is that it allows
+fine-grained epoch assignment.
+In this approach an Attester's signature spans the nonce as well as an
+Attester's trusted timetick counter(s).
```
dthaler, line 888: For parity with previous subsection, this would then require adding text saying "This typically requires additional claims about the Attester's timetick counter mechanism in order to provide such assurance."

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+
+## Interval Timekeeping using Nonces
+
+A third approach is a variant of the first; the difference is that it allows
+fine-grained epoch assignment.
+In this approach an Attester's signature spans the nonce as well as an
+Attester's trusted timetick counter(s). 
+Signatures across subsequent Evidence which embeds incrementing values of
+the trusted timetick counter(s) can be used to establish the time interval
+which has elasped since the nonce was received.
```
dthaler, line 891: I don't yet follow why the time interval since nonce was received is interesting for anything.  Explain, or provide a forward reference if it's later.

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+
+## Distributed Interval Timekeeping using Local Clocks
+
+A fourth approach places the onus of timekeeping on all involved parties,
+but does not require the parties' clocks to be sychronized.
+A centralized entity is required to distribute non-repeatable values, which are referred to as
+Handles, to all involved parties in periodic intervals.
+Each Handle has a validity-span with an epoch that starts at the local time of
```
dthaler, line 899: The term "validity-span" is not a normal term as far as I know

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+involved parties at which a distributed handle has been received.
+While only the source of the periodically sent handle must track the time interval, Attesters should also track is to avoid creating invalid Evidence when it can be avoided.
```
dthaler, line 901: can't parse "track is to avoid"

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+
+A third approach is a variant of the first; the difference is that it allows
+fine-grained epoch assignment.
+In this approach an Attester's signature spans the nonce as well as an
+Attester's trusted timetick counter(s). 
+Signatures across subsequent Evidence which embeds incrementing values of
+the trusted timetick counter(s) can be used to establish the time interval
+which has elasped since the nonce was received.
+
+## Distributed Interval Timekeeping using Local Clocks
```
dthaler, line 893: I can't figure out how/why this is any different architecturally from nonces.  To me, the handle is just a nonce as far as I can tell from the below discussion.

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+of Verifiers the time window of validity expired and the Evidence
+cannot be considered fresh anymore.
+
+## Claim Validity
+
+In all approaches, it is important to note that the actual values in Claims
+might have been generated long before the Claims are signed.
+If so, it is the
+signer's (the Attestation Environment's) responsibility to ensure that the values are still correct when they
+are signed.
```
dthaler, line 915: I'm not sure I agree with this.  Shouldn't it be covered by the policy?  If an attester can only get a stale value and can't vouch for it not having been changed since, then putting that fact in the Evidence allows the verifier to make a decision appropriately.

mcr:
> This commit slightly changes the Freshness section to reflect the
> outcome of the discussion around PR #87. (Hopefully I've managed to
> capture the gist of Henk's thinking without disrupting too much the
> existing flow.)

So, if #87 is dead, can you close it?

draft-ietf-rats-architecture.md
William-PanWei: +1, I agree with Dave

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
+## Distributed Interval Timekeeping using Local Clocks
+
+A fourth approach places the onus of timekeeping on all involved parties,
+but does not require the parties' clocks to be sychronized.
+A centralized entity is required to distribute non-repeatable values, which are referred to as
+Handles, to all involved parties in periodic intervals.
+Each Handle has a validity-span with an epoch that starts at the local time of
+involved parties at which a distributed handle has been received.
+While only the source of the periodically sent handle must track the time interval, Attesters should also track is to avoid creating invalid Evidence when it can be avoided.
+Distributed handles are used in place of a nonces for Evidence generation.
```
William-PanWei, line 902:
```suggestion
Distributed handles are used in place of nonces for Evidence generation.
```

draft-ietf-rats-architecture.md
William-PanWei: It seems like that the handle is also a nonce, and the difference is that this nonce is from a third party (a centralized entity), not the Verifier or Relying Party.
I'm afraid this approach is more complex because you need to consider the time sync among three entities. For example, the time of receiving the handle among different entities may be different, therefore the handle expired time will be different. If the handle on the Verifier expires earlier than the Attester, the handle may already be expired while Verifier receives the Evidence from Attester.

draft-ietf-rats-architecture.md
```diff
@@ -961,23 +1001,27 @@ Carsten Bormann provided many of the motivational building blocks with respect t
...
-| NR | Nonce relayed               | The nonce is relayed to an Attester by enother entity
-| EG | Evidence generation         | An Attester collects claims and generates Evidence
-| ER | Evidence relayed            | A Relying Party relays Evidence to a Verifier
-| RG | Result generation           | A Verifier appraises Evidence and generates an Attestation Result
-| RR | Result relayed              | A Relying Party relays an Attestation Result to a Relying Party
-| RA | Result appraised            | The Relying Party appraises Attestation Results
+| VG | Value generation            | A value to appear in a Claim was created.
+| AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
+| CC | Claim Collection            | An Attesting Environment collects a new/changed Claim value to appear in Evidence.
+| HD | Handle distribution         | A centrally generated identifier for time-bound recentness across a domain of devices is successfully received by Attesters (and Verifiers).
```
William-PanWei, line 1012: I think these new-added events had been questioned before, that the examples below don't use them.

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
-signer's responsibility to ensure that the values are still correct when they
-are signed.  For example, values might have been generated at boot, and then
-used in claims as long as the signer can guarantee that they cannot have changed
-since boot.
+# Freshness
+
+A remote entity (Verifier or Relying Party) may need to learn the point in
+time (i.e., the "epoch") an Evidence or Attestation Result have been
+produced.
+This is essential in deciding whether the included Claims can be considered fresh, meaning they still reflect the latest state of the Attester, and that any Attestation Result was generated using the latest Appraisal Policy for Evidence.
```
mcr, line 850:
```suggestion
This is essential in deciding whether the included Claims and their values can be considered fresh, meaning they still reflect the latest state of the Attester, and that any Attestation Result was generated using the latest Appraisal Policy for Evidence.
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+Freshness is assessed based on a policy defined by the consuming entity,
+Verifier or Relying Party, that compares the estimated epoch against an
+"expiry" threshold defined locally to that policy.
+There is, however, always a race condition possible in that the state of the Attester, and the Appraisal Policy for Evidence, might change immediately after the Evidence or Attestation Result was generated.
+The goal is merely to narrow their recentness to something the Verifier (for Evidence) or Relying Party (for Attestation Result) is willing to accept.
+Freshness is a key component  for enabling caching and reuse of both Evidence and Attestation Results, which is especially valuable in cases where their computation uses a substantial part of the energy budget (e.g., in constrained devices).
```
mcr, line 857:
```suggestion
Freshness is a key component  for enabling caching and reuse of both Evidence and Attestation Results, which is especially valuable in cases where their computation uses a substantial part of the resource budget (e.g., energy in constrained devices).
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+There are two common approaches for determining the epoch of an Evidence or Attestation Result.
+
+<!-- Explicit Timekeeping using Synchronized Clocks -->
+
+The first approach is to rely on synchronized and trustworthy clocks, and include a signed timestamp (see {{?I-D.birkholz-rats-tuda}}) along with the Claims in the Evidence or Attestation Result.
+This allows fine-grained epoch assignment, even on per-Claim basis if wanted, as well as the ability to distinguish time of collection from time of creation of Claims.
```
mcr, line 864:
```suggestion
This allows fine-grained epoch assignment, even on a per-Claim basis if wanted, as well as the ability to distinguish time of collection from time of creation of Claims.
```

dthaler, line 864:
```suggestion
Timestamps can also be added on a per-Claim basis if wanted, to allow distinguishing the time of creation of Evidence from the time that a specific Claim value was generated.
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+Clocks trustworthiness typically requires additional Claims about the signer's time synchronization mechanism.
+
+<!-- Implicit Timekeeping using Nonces -->
+
+A second approach places the onus of timekeeping solely on the remote entity (e.g., the Verifier for Evidence, or the Relying Party for an Attestation Result).
+This might be an appropriate choice in case the Attester does not have a reliable clock or time synchronisation is otherwise impaired.
+In this approach, a non-predictable nonce is generated by the remote entity, and the nonce is then signed and included along with the Claims in the Evidence or Attestation Result.
```
dthaler, line 871:
```suggestion
In this approach, a non-predictable nonce is sent by the remote entity, and the nonce is then signed and included along with the Claims in the Evidence or Attestation Result.
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+After checking the sent and received nonces are the same, the remote entity knows that the Claims were signed after the nonce was generated, effectively locking its clock to the produced payload.
```
dthaler, line 872:
```suggestion
After the remote entity checks that the sent and received nonces are the same, the remote entity knows that the Claims were signed after the nonce was generated.
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+This allows associating a "rough" epoch to the Evidence or Attestation Result.
+In this case the epoch is said to be rough because:
+
+* it is always bound to the set of Claims as a whole, as opposed to having Claim being separately annotated, and
+* it is not possible to distinguish between time of collection from the time of creation of the Claims set.
```
dthaler, line 877:
```suggestion
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+The two basic mechanisms presented above can be extended and combined in more advanced schemes.
+For example, if the clocks are trustworthy but not synchronized, first a nonce-based exchange may be used to determine the (rough) time offset between the involved peers, followed by any number of timestamp based exchanges.
```
dthaler, line 880:
```suggestion
For example, if clocks exist and are considered trustworthy but are not synchronized, first a nonce-based exchange may be used to determine the (rough) time offset between the involved peers, followed by any number of timestamp based exchanges.
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+<!-- Explicit Timekeeping using Synchronized Clocks -->
+
+The first approach is to rely on synchronized and trustworthy clocks, and include a signed timestamp (see {{?I-D.birkholz-rats-tuda}}) along with the Claims in the Evidence or Attestation Result.
+This allows fine-grained epoch assignment, even on per-Claim basis if wanted, as well as the ability to distinguish time of collection from time of creation of Claims.
+Clocks trustworthiness typically requires additional Claims about the signer's time synchronization mechanism.
+
+<!-- Implicit Timekeeping using Nonces -->
+
+A second approach places the onus of timekeeping solely on the remote entity (e.g., the Verifier for Evidence, or the Relying Party for an Attestation Result).
```
mcr, line 869:
```suggestion
A second approach places the onus of timekeeping solely on the appraising entity (e.g., the Verifier for Evidence, or the Relying Party for an Attestation Result).
```

draft-ietf-rats-architecture.md
mcr, line 871:
```suggestion
In this approach, a non-predictable nonce is generated by the appraising entity, and the nonce is then signed and included along with the Claims in the Evidence or Attestation Result.
```

draft-ietf-rats-architecture.md
mcr, line 872:
```suggestion
After checking the sent and received nonces are the same, the appraising entity knows that the Claims were signed after the nonce was generated, effectively locking its clock to the produced payload.
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+Timestamps can also be added on a per-Claim basis if wanted, to allow distinguishing the time of creation of Evidence from the time that a specific Claim value was generated.
+Clocks trustworthiness typically requires additional Claims about the signer's time synchronization mechanism.
+
+<!-- Implicit Timekeeping using Nonces -->
+
+A second approach places the onus of timekeeping solely on the appraising entity (e.g., the Verifier for Evidence, or the Relying Party for an Attestation Result).
+This might be an appropriate choice in case the Attester does not have a reliable clock or time synchronisation is otherwise impaired.
+In this approach, a non-predictable nonce is generated by the appraising entity, and the nonce is then signed and included along with the Claims in the Evidence or Attestation Result.
+After checking the sent and received nonces are the same, the appraising entity knows that the Claims were signed after the nonce was generated, effectively locking its clock to the produced payload.
```
dthaler, line 872:
```suggestion
After checking the sent and received nonces are the same, the appraising entity knows that the Claims were signed after the nonce was generated.
```

draft-ietf-rats-architecture.md
thomas-fossati: I don't think we should ditch this.

The essential premise is on line 869. In particular, by using "solely" I wanted to discuss the case where the only clock that the appraiser can trust is its own.  The intent of line 870 is just to illustrate two common instances where the appraiser can't rely on the peer's clock.

The overall premise for this whole paragraph is that appraiser does _not_ consider any time assertion made the peer, and to determine what kind of information can be used in this case to evaluate freshness.

Clearly, once the peer's clock becomes usable - as in the example at line 880 - then finer-grained timestamping can become input to the evaluation of the freshness policy as well.

Hopefully I have made my intention clear and the ambiguities in the existing text can be fixed accordingly :-)

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
-claims in the Evidence or Attestation Result, so that the remote entity knows
-that the claims were signed at that time, as long as it has some assurance that
-the timestamp is correct.  This typically requires additional claims about
-the signer's time synchronization mechanism in order to provide such assurance.
-
-In either approach, it is important to note that the actual values in claims
-might have been generated long before the claims are signed.  If so, it is the
-signer's responsibility to ensure that the values are still correct when they
-are signed.  For example, values might have been generated at boot, and then
-used in claims as long as the signer can guarantee that they cannot have changed
```
nedmsmith, line 877: "could not have"

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+There is, however, always a race condition possible in that the state of the Attester, and the Appraisal Policy for Evidence, might change immediately after the Evidence or Attestation Result was generated.
+The goal is merely to narrow their recentness to something the Verifier (for Evidence) or Relying Party (for Attestation Result) is willing to accept.
+Freshness is a key component  for enabling caching and reuse of both Evidence and Attestation Results, which is especially valuable in cases where their computation uses a substantial part of the resource budget (e.g., energy in constrained devices).
+
+There are two common approaches for determining the epoch of an Evidence or Attestation Result.
+
+<!-- Explicit Timekeeping using Synchronized Clocks -->
+
+The first approach is to rely on synchronized and trustworthy clocks, and include a signed timestamp (see {{?I-D.birkholz-rats-tuda}}) along with the Claims in the Evidence or Attestation Result.
+Timestamps can also be added on a per-Claim basis if wanted, to allow distinguishing the time of creation of Evidence from the time that a specific Claim value was generated.
```
nedmsmith, line 864: "also", "if wanted", "of Evidence", "value" are not needed.
"allow distinguishing" change to "distinguish"

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+Clocks trustworthiness typically requires additional Claims about the signer's time synchronization mechanism.
```
nedmsmith, line 865: "The clock's trustworthiness" - grammar.

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+<!-- Implicit Timekeeping using Nonces -->
+
+A second approach places the onus of timekeeping solely on the appraising entity (e.g., the Verifier for Evidence, or the Relying Party for an Attestation Result).
```
nedmsmith, line 869: "the Verifier (for Evidence), or the Relying Party (for Attestation Results)."

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+This might be an appropriate choice in case the Attester does not have a reliable clock or time synchronisation is otherwise impaired.
+In this approach, a non-predictable nonce is generated by the appraising entity, and the nonce is then signed and included along with the Claims in the Evidence or Attestation Result.
+After checking the sent and received nonces are the same, the appraising entity knows that the Claims were signed after the nonce was generated, effectively locking its clock to the produced payload.
+This allows associating a "rough" epoch to the Evidence or Attestation Result.
```
nedmsmith, line 873: Replace "This" with "Timekeeping" ?

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+In this case the epoch is said to be rough because:
+
+* it is always bound to the set of Claims as a whole, as opposed to having Claim being separately annotated, and
```
nedmsmith, line 876: * The epoch applies to the entire claim set instead of a more granular association, and

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+* it is not possible to distinguish between time of collection from the time of creation of the Claims set.
```
nedmsmith, line 877: * The time between the creation of Claims and the collection of Claims is indistinguishable.

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+The two basic mechanisms presented above can be extended and combined in more advanced schemes.
```
nedmsmith, line 879: "Implicit and explicit timekeeping can be combined into hybrid mechanisms.

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+For example, if the clocks are trustworthy but not synchronized, first a nonce-based exchange may be used to determine the (rough) time offset between the involved peers, followed by any number of timestamp based exchanges.
+In another scenario where a broadcast channel is shared by all the RATS entities (Attesters, Verifiers and Relying Parties), the nonce-based approach may be used to lock all parties to the same timeline without requiring synchronized clocks by having a central entity emit nonces at regular intervals and have the "current" nonce included in the produced Evidence or Attestation Result.
```
nedmsmith, line 881: "Second, a broadcast channel shared by all Roles (Attester, Verifier, etc.), the nonce-based approach..."

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+
+Whatever the approach, it is important to note that the actual values in Claims might have been generated long before the Claims are signed.
```
nedmsmith, line 883: "It is important..."

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,47 @@ a different format.
...
+If so, it is the signer's responsibility to ensure that the values are still correct when they are signed.
+For example, values might have been generated at boot, and then used in Claims as long as the signer can guarantee that they cannot have changed since boot.
```
nedmsmith, line 885: This is not an example imo. "For example, values generated at boot time might have been saved to secure storage until network connectivity is established to the remote Verifier and a nonce is obtained."

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: Not sure: "this" is a bit more than just "timekeeping", it refers to "an approach that places the onus of timekeeping solely on the appraising entity".

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: Not sure I understand this suggestion.  If the current text isn't good as-is, we may simplify it as "When a broadcast channel is shared by all Roles, the nonce-base approach [...]"

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: fixed in c215569

draft-ietf-rats-architecture.md
thomas-fossati: This change has been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: This change has been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: This change has been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: This change has been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: This change has been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: Yes.  All of these changes have now been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: This text doesn't exist anymore.

draft-ietf-rats-architecture.md
thomas-fossati: This has been addressed using Ned's suggestion in c215569

draft-ietf-rats-architecture.md
thomas-fossati: This text doesn't exist anymore.

draft-ietf-rats-architecture.md
thomas-fossati: This text doesn't exist anymore.

draft-ietf-rats-architecture.md
thomas-fossati: This text doesn't exist anymore.

draft-ietf-rats-architecture.md
thomas-fossati: This text doesn't exist anymore.

draft-ietf-rats-architecture.md
thomas-fossati: This text doesn't exist anymore.

thomas-fossati:
> > This commit slightly changes the Freshness section to reflect the
> > outcome of the discussion around PR #87. (Hopefully I've managed to
> > capture the gist of Henk's thinking without disrupting too much the
> > existing flow.)
> 
> So, if #87 is dead, can you close it?

Done.

draft-ietf-rats-architecture.md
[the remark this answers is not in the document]
thomas-fossati: This text doesn't exist anymore.

draft-ietf-rats-architecture.md
```diff
@@ -373,60 +378,78 @@ Attesting Environments are designed specifically with claims collection in mind.
...
-in order to create believable attestation Evidence.
-
-{{layered}} depicts an example of a device that includes (A) a BIOS stored
-in read-only memory in this example, (B) an updatable bootloader, and (C)
-an operating system kernel.
+Attester -- in itself -- is composed and therefore has to conduct the Claims collection
+in order to create believable Attestation Evidence.
+The following example is intended to illustrate this composition:
 
+A very common example is elaborated on to illustrate Layered Attestation.
```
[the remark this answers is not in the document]
thomas-fossati: This has been reverted.

draft-ietf-rats-architecture.md
```diff
@@ -844,40 +835,89 @@ a different format.
...
-Result are fresh, meaning that there is some degree of assurance
-that they still reflect the latest state of the Attester, and that any Attestation
-Result was generated using the latest Appraisal Policy for Evidence.
+It is important to prevent old Evidence or old Attestation Results that are no longer correct to be assessed as valid. Replay attacks are an important reason to do that. More importantly, the validity of Conceptual Messages, such as Evidence or Attestation Results, are the basis for trustworthy remote attestation procedures, in general.
+Some mechanism of ensuring that Evidence and Attestation
+Results are actually fresh are required. In essence, there has to be some degree of assurance
+that they still reflect the latest operational state of the Attester and that, for example, Attestation
+Results were generated using the latest Appraisal Policy for Evidence.
 There is, however, always a race condition possible in that the
 state of the Attester, and the Appraisal Policy for Evidence,
```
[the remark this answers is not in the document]
thomas-fossati: This text has been reverted.

draft-ietf-rats-architecture.md
thomas-fossati: This whole chunk of text is gone.

draft-ietf-rats-architecture.md
thomas-fossati: This chunk of text is gone.

draft-ietf-rats-architecture.md
thomas-fossati: Fixed in 34c9443

draft-ietf-rats-architecture.md
thomas-fossati: Fixed in 34c9443

draft-ietf-rats-architecture.md
thomas-fossati: Fixed in 35dc70a

draft-ietf-rats-architecture.md
thomas-fossati: Ned, you had another comment on this same sentence, which replaced it altogether.  Since your later suggestion has been already incorporated (c215569), I'm closing this.

draft-ietf-rats-architecture.md
thomas-fossati: Done in b0488b2 for the whole Freshness section.

draft-ietf-rats-architecture.md
nedmsmith: Maybe substitute your text in place of 'this' since the pronoun reference seems ambiguous?

draft-ietf-rats-architecture.md
nedmsmith: Trying to get the signposting to match up. In line 880 signposting uses "first" and in 881 signposting changes to "another". Normally, "second" follows "first" - it is just style and I don't feel strongly about it.

thomas-fossati: Line 880 and 881 are two separate examples.

The example at line 880 is split in two phases: first [...], followed by [...].

I will remove "first" since it isn't necessary and I suspect it's probably bad English if it got you confused.

draft-ietf-rats-architecture.md
thomas-fossati: That seems a bit redundant.  I'd like to achieve some kind of compression - obviously without introducing ambiguity.  Let me see if I can concoct something decent.

draft-ietf-rats-architecture.md
thomas-fossati: Ned: please check the proposal in 20e3d7

draft-ietf-rats-architecture.md
thomas-fossati: Ned: please check the proposal in 20e3d7

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,73 @@ a different format.
...
+Claims in the Evidence or Attestation Result.  Timestamps can be added on a
+per-Claim basis, to distinguish the time of creation from the time that a
+specific Claim was generated.  The clock's trustworthiness typically requires
+additional Claims about the signer's time synchronization mechanism.
+
+<!-- Implicit Timekeeping using Nonces -->
+
+A second approach places the onus of timekeeping solely on the appraising
+entity, e.g., the Verifier (for Evidence), or the Relying Party (for
+Attestation Results), and might be suitable in case the Attester does not have
```
William-PanWei, line 882: I think 'the Attester' here can't cover the case where the Relying Party sends the nonce to the Verifier for retrieving the Attestation Result. Maybe change 'the Attester' to 'the producing entity'?

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,73 @@ a different format.
...
+
+There are two common approaches for determining the epoch of an Evidence or
+Attestation Result.
+
+<!-- Explicit Timekeeping using Synchronized Clocks -->
+
+The first approach is to rely on synchronized and trustworthy clocks, and
+include a signed timestamp (see {{?I-D.birkholz-rats-tuda}}) along with the
+Claims in the Evidence or Attestation Result.  Timestamps can be added on a
+per-Claim basis, to distinguish the time of creation from the time that a
```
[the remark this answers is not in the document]
thomas-fossati: > Does 'the time of creation' mean the time of the creation of the Evidence or Attestation Result?

Yes.

> More explicit would be better.

OK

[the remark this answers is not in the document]
thomas-fossati: fixed in 8232e0a

draft-ietf-rats-architecture.md
thomas-fossati: 1. You are right in that it is not limited to the Attester.  I wanted to use the it as the typical example because it's more likely that a constrained Attester doesn't have a very reliable clock.  What about: `and might be suitable, for example, in case the Attester does not have [...]`?

2. The current paragraph looks wrong to me.  When it says:
```
[...] solely on the appraising entity, e.g., the Verifier (for Evidence),
or the Relying Party (for Attestation Results),`
```
should instead say:
```
[...] solely on the remote attestation protocol initiator, e.g., the
Relying Party (for Evidence or Attestation Results) and the Attester
(for Attestation Results).
```

draft-ietf-rats-architecture.md
```diff
@@ -842,40 +842,75 @@ a different format.
...
+per-Claim basis, to distinguish the time of creation of Evidence or Attestation
+Result from the time that a specific Claim was generated.  The clock's
+trustworthiness typically requires additional Claims about the signer's time
+synchronization mechanism.
+
+<!-- Implicit Timekeeping using Nonces -->
+
+A second approach places the onus of timekeeping solely on the appraising
+entity, e.g., the Verifier (for Evidence), or the Relying Party (for
+Attestation Results), and might be suitable in case the Attester does not have
```
mcr, line 883:
```suggestion
A second approach places the onus of timekeeping solely on the appraising
entity, e.g., the Verifier (for Evidence), or the Relying Party (for
Attestation Results), and might be suitable, for example, in case the Attester does not have
```

PR #93, opened 2020-06-05
Authors don't need to be in acknowledgments section

dthaler:
Wei Pan was added to the author list but wasn't removed from the
Acknowledgements section at the time so now appears in both places.
This PR corrects that.


henkbirkholz:
Yeah. I think this is yet another artifact and Dave is absolutely right here. Wei is already an "acknowledged editor" and therefore obviously does not has to be included in a potential ack section, apparently.

Issue #65, opened 2020-03-10
More thorough definition of Endorser or Endorsement


laurencelundblade:
Here's some possible text.

Endorser
The primary function of an Endorser is to endorse an Attester so that a Verifier can know to trust that Attester. The Endorser does this by creating an Endorsement and conveying it securely to the Verifier. The primary basis for a Verifier trusting an Attester is the Endorsement.

There is thus an existential relationship between Attesters and Endorsers.  An Attester with no Endorsement is not worth very much since there is no good basis by which a Verifier can trust it so every Attester needs an Endorser. The Endorser’s primary function is to endorse Attesters so an Endorser doesn’t have anything to do without some Attesters to endorse.

A Verify cannot do its job without Endorsements. It must have an Endorsement for each Attester from which it received Attestation Evidence. Without this, the Attester is not trusted and there is no basis for trusting the Evidence it produces.

There is a trust triangle between the Endorser, Attester and Verifier. The Endorser tells the Verifier that an Attester is trustworthy.

Endorsement
Ultimately an Endorsement must be realized as some input bytes to a Verifier and they must be conveyed securely from the Endorser to the Verifier. RATS architecture does not impose any limit or restriction on the formats of these bytes or the conveyance mechanism.  In practice they are expected to vary quite widely.

In most cases the Endorsement will include the key material that is used to verify signatures on the Evidence. Since the Evidence will contain a nonce, this signature verification serves to authenticate the attester. The Endorsement may also contain information about the Attester such as its type, version, manufacturer and such.

In one example, the Endorser and the Verifier are the same legal organization, so secure conveyance is entirely an internal issue for the organization. It may consist of one employee copying a set of public keys from computer in the company to another.

In another example, the Endorsement may take the form of a PKI hierarchy. The Endorser conveys a root certificate to the Verifier. The Endorser also creates per-attester certificates under that root. Those per-attester certificates are programmed into Attesters. The Attester sends them to the  Verifier attached to the Evidence. There may be several levels of certificates in the PKI hierarchy. The Verifier walks the certificate chain to determine trust.  Note in the case the abstract notion of Endorsement comes to the Verifier in two parts, one in an out-of-band secure conveyance or a root certificate and the other as the signed (self-securing) certificates from the Attester.

Unendorsed Attesters
[...]

PR #94, opened 2020-06-08
More description of Endorsements

laurencelundblade:
Addresses issue #65

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -734,47 +734,27 @@ or interval when changes in operational status, health, and so forth occur.
 
 ## Endorsements
 
-Endorsements are essential input to the Verifier that allows them to
-trust Attesters and in turn the Evidence produced by Attesters.
-Without endorsements Verifiers cannot do their job.
+An Endorsement input to a Verifier is always required. An Endorsement is what allows a Verifier to trust an Attester and in turn to trust the Evidence produced by an Attester. Without Endorsements a Verifier cannot function.
+Said another way, an Attester for which there is no Endorsement is not useful since there is no good basis by which a Verifier can trust it.
 
-An Attester for which there is no Endorsement is not very useful since
-there is no good basis by which a Verifier can trust it.
-Basically, every Attester needs an Endorsement.
+One of the Endorsements for an Attester will just about always contain cryptographic key material.
+This key material will correspond to key material that is put in the Attester.
+The key material is the basis for the Verifier trusting the Attester.
+This architecture does not dictate any particular cryptographic protocol for establishing this trust, hence the key material can be symmetric or asymmetric, a single key, multiple keys, key hierarchies, X.509 cert or chains or other.
 
-An Endorsement usually contains some key material, often a public key,
-that corresponds to a private key that is in the Attester.
-This key material is used to prove the identity of the Attester to
-the Verifier.
-It is also is usually used to sign and verify the Evidence.
-The key material may be as simple as one key pair, with the public key
-in the Endorsement and the private key in the Attester, or it may be
-more elaborate using an X.509 hierarchy or such.
-It may also be symmetric key material. This architecture allows any
-type of key material and scheme.
+Endorsements are produced by an Endorser.
+The Endorser is often the manufacturer of the device that contains the Attester, but not necessarily so.
+For example, it could be an OS vendor that supplies Attester key material to the device/Attester manufacturer.
 
-Endorsements may also contain Claims about the Attester.
-These are claims about the Attester that are static, that don’t vary
-with the state of the device or the Attester.  
-Examples include the manufacturer and model.
-They may be used along with Appraisal Policy to evaluate the Attester
-and Evidence it produces.
-They may also be passed through into Attestation Results for use by
-the Relying Party or both.
+An Endorsement must be conveyed to the Verifier in a secure way so that the Verifier does not accept fake or false Endorsements that would result in trusting Attesters and Evidence that should not be trusted.
+This architecture does not dictate any particular means of securing this conveyance hence it might be a TLS connection, some sort of signed data, physically transferring the data or other.
 
-Endorsements are usually produced by the manufacturer of the Attester.
-Sometimes the manufacturer is called the Endorser. 
-Sometimes there may be many party involved in producing the device,
-the Attester and the Endorsement. For example, an OS vendor might be
-the one producing the Endorsements and supplying key material to the
-actual manufacturer.
+Endorsements may contain known-good or reference values that are used by the Verifier to evaluate Evidence in the application of the Appraisal Process.
 
-The endorsement must be conveyed to the Verifier in a secure way. If
-not, the Verifier could be told to trust any Attester.
-The specific security mechanism for this is not specified by this
-architecture. Some examples might be a common trusted X.509 root
-certificate used with TLS, a private connection or some out-of-band
-delivery of key material.
+Endorsements may contain implicit Claims about the Attester.
+These are Claims about the device/Attester that are static, that don’t vary with the state of the device or the Attester.
+These implicit static Claims are usually passed on to the Relying Party in Attestation Results.
+The name of the manufacturer and model number are examples of implicit static Claims.
 
 Thus, when Evidence and such an Endorsement
 are used together, an appraisal procedure can be conducted based on Appraisal Policies that may not be specific to the
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
 
 ## Endorsements
 
-An Endorsement is a secure statement that some entity (e.g., a manufacturer) vouches for the integrity of the
-device's signing capability.  For example, if the signing capability is in hardware, then
-an Endorsement might be a manufacturer certificate that signs a public key whose corresponding
-private key is only known inside the device's hardware.  Thus, when Evidence and such an Endorsement
+Endorsements are essential input to the Verifier that allows them to
+trust Attesters and in turn the Evidence produced by Attesters.
+Without endorsements Verifiers cannot do their job.
+
+An Attester for which there is no Endorsement is not very useful since
+there is no good basis by which a Verifier can trust it.
+Basically, every Attester needs an Endorsement.
+
+An Endorsement usually contains some key material, often a public key,
+that corresponds to a private key that is in the Attester.
+This key material is used to prove the identity of the Attester to
+the Verifier.
+It is also is usually used to sign and verify the Evidence.
+The key material may be as simple as one key pair, with the public key
+in the Endorsement and the private key in the Attester, or it may be
+more elaborate using an X.509 hierarchy or such.
+It may also be symmetric key material. This architecture allows any
+type of key material and scheme.
+
+Endorsements may also contain Claims about the Attester.
+These are claims about the Attester that are static, that don’t vary
+with the state of the device or the Attester.  
+Examples include the manufacturer and model.
+They may be used along with Appraisal Policy to evaluate the Attester
+and Evidence it produces.
+They may also be passed through into Attestation Results for use by
+the Relying Party or both.
+
+Endorsements are usually produced by the manufacturer of the Attester.
+Sometimes the manufacturer is called the Endorser. 
+Sometimes there may be many party involved in producing the device,
+the Attester and the Endorsement. For example, an OS vendor might be
+the one producing the Endorsements and supplying key material to the
+actual manufacturer.
+
+The endorsement must be conveyed to the Verifier in a secure way. If
+not, the Verifier could be told to trust any Attester.
+The specific security mechanism for this is not specified by this
+architecture. Some examples might be a common trusted X.509 root
+certificate used with TLS, a private connection or some out-of-band
+delivery of key material.
+
+Thus, when Evidence and such an Endorsement
 are used together, an appraisal procedure can be conducted based on Appraisal Policies that may not be specific to the
 device instance, but merely specific to the manufacturer providing the Endorsement. For example,
 an Appraisal Policy might simply check that devices from a given manufacturer have information
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -747,6 +747,8 @@ The Endorser is often the manufacturer of the device that contains the Attester,
 For example, it could be an OS vendor that supplies Attester key material to the device/Attester manufacturer.
 
 An Endorsement must be conveyed to the Verifier in a secure way so that the Verifier does not accept fake or false Endorsements that would result in trusting Attesters and Evidence that should not be trusted.
+Signing is used in some Endorsement formats, but not in all.
+When signing is not used, some other security mechanism is necessary.
 This architecture does not dictate any particular means of securing this conveyance hence it might be a TLS connection, some sort of signed data, physically transferring the data or other.
 
 Endorsements may contain known-good or reference values that are used by the Verifier to evaluate Evidence in the application of the Appraisal Process.


draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
 
 ## Endorsements
 
-An Endorsement is a secure statement that some entity (e.g., a manufacturer) vouches for the integrity of the
-device's signing capability.  For example, if the signing capability is in hardware, then
-an Endorsement might be a manufacturer certificate that signs a public key whose corresponding
-private key is only known inside the device's hardware.  Thus, when Evidence and such an Endorsement
+Endorsements are essential input to the Verifier that allows them to
```
nedmsmith, line 737: singular / plural agreement

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+An Endorsement usually contains some key material, often a public key,
+that corresponds to a private key that is in the Attester.
+This key material is used to prove the identity of the Attester to
+the Verifier.
+It is also is usually used to sign and verify the Evidence.
+The key material may be as simple as one key pair, with the public key
+in the Endorsement and the private key in the Attester, or it may be
+more elaborate using an X.509 hierarchy or such.
+It may also be symmetric key material. This architecture allows any
+type of key material and scheme.
```
nedmsmith, line 754: Not sure lines 750 - 754 are necessary.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+
+Endorsements may also contain Claims about the Attester.
```
nedmsmith, line 756: "Endorsements may contain reference Claims and invariant Claims. Reference Claims are expected values that can be compared to actual values supplied as Evidence. Invariant Claims are static values that always apply regardless of Attester operational states." (replaces 756 - 758)

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+These are claims about the Attester that are static, that don’t vary
+with the state of the device or the Attester.  
+Examples include the manufacturer and model.
+They may be used along with Appraisal Policy to evaluate the Attester
```
nedmsmith, line 760: I don't think 760 - 761 are needed.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+and Evidence it produces.
+They may also be passed through into Attestation Results for use by
```
nedmsmith, line 762: 762 - 763 is not stated correctly. Claims included in Attestation Results are no longer considered Endorsements. The section on Attestation Results should suffice as a definition of Attestation Results.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+the Relying Party or both.
+
+Endorsements are usually produced by the manufacturer of the Attester.
```
nedmsmith, line 765: "usually" is incorrect stereotyping. Not sure 765 is adding useful information.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+Sometimes the manufacturer is called the Endorser. 
+Sometimes there may be many party involved in producing the device,
+the Attester and the Endorsement. For example, an OS vendor might be
+the one producing the Endorsements and supplying key material to the
+actual manufacturer.
```
nedmsmith, line 770: Lines 766 - 770 are exemplary in nature, but reads somewhat prescriptive. Suggest we re-look at "Endorser" section to see if it captures the use case intuition.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+
+The endorsement must be conveyed to the Verifier in a secure way. If
+not, the Verifier could be told to trust any Attester.
+The specific security mechanism for this is not specified by this
+architecture. Some examples might be a common trusted X.509 root
+certificate used with TLS, a private connection or some out-of-band
+delivery of key material.
```
nedmsmith, line 777: There is a section on message conveyance and on trust model. If these sections are not clear enough, maybe changes should be considered there?

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+
+Thus, when Evidence and such an Endorsement
```
nedmsmith, line 779: I'm not sure what this sentence is saying.

draft-ietf-rats-architecture.md
dthaler: I disagree with this phrasing.   Endorsements are essential for scalability.  If a Verifier owner has some means to establish trust in each Attester's hardware individually then it need not use an endorsement.  But typically that's not scalable.   Thus I prefer the original wording.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
 
 ## Endorsements
 
-An Endorsement is a secure statement that some entity (e.g., a manufacturer) vouches for the integrity of the
-device's signing capability.  For example, if the signing capability is in hardware, then
-an Endorsement might be a manufacturer certificate that signs a public key whose corresponding
-private key is only known inside the device's hardware.  Thus, when Evidence and such an Endorsement
+Endorsements are essential input to the Verifier that allows them to
+trust Attesters and in turn the Evidence produced by Attesters.
+Without endorsements Verifiers cannot do their job.
```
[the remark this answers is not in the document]
dthaler: I disagree with both the proposed line 739 and Ned's statement.  Endorsements don't contain "expected values", those are obtained (if at all) in addition to the "endorsement" per se, not as part of them.  (My point is just about terminology/definition, not about the protocol where they may typically come in the same protocol as endorsements.)

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+
+An Attester for which there is no Endorsement is not very useful since
+there is no good basis by which a Verifier can trust it.
+Basically, every Attester needs an Endorsement.
```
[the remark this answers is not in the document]
dthaler: Agree they're redundant, and I disagree with it for the same reason :)

draft-ietf-rats-architecture.md
dthaler: I agree with Ned this is too much protocol details, and may not apply to all implementations so isn't really architectural I think.  Same goes for lines 747-749 I would say.

draft-ietf-rats-architecture.md
dthaler: Agree with Ned.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
...
+the Relying Party or both.
+
+Endorsements are usually produced by the manufacturer of the Attester.
+Sometimes the manufacturer is called the Endorser. 
+Sometimes there may be many party involved in producing the device,
+the Attester and the Endorsement. For example, an OS vendor might be
+the one producing the Endorsements and supplying key material to the
+actual manufacturer.
+
+The endorsement must be conveyed to the Verifier in a secure way. If
```
dthaler, line 772: This implies conveyance via a secure protocol, which I disagree is a must.  As long as endorsements are signed, how they're conveyed is irrelevant... conveyance via an insecure protocol is fine.

draft-ietf-rats-architecture.md
laurencelundblade: The architecture needs to include the Verifier knows to trust the Attester, how it gets the key material on which the trust is based.  Since the only inputs are Endorsements and Policy, Endorsement seems the better line up.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,49 @@ or interval when changes in operational status, health, and so forth occur.
 
 ## Endorsements
 
-An Endorsement is a secure statement that some entity (e.g., a manufacturer) vouches for the integrity of the
-device's signing capability.  For example, if the signing capability is in hardware, then
-an Endorsement might be a manufacturer certificate that signs a public key whose corresponding
-private key is only known inside the device's hardware.  Thus, when Evidence and such an Endorsement
+Endorsements are essential input to the Verifier that allows them to
+trust Attesters and in turn the Evidence produced by Attesters.
```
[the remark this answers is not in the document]
laurencelundblade: Well, yes, but if you don't trust the Attester to some degree, the Evidence is worthless. The sentence does depict that.

draft-ietf-rats-architecture.md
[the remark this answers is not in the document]
laurencelundblade: The Verifier has to have key material. It may also get implied or inferred claims, things it knows about the Attester that the Attester didn't tell it directly. From discussions with Henk and such, I believe both the key material and implied claims come in the form of Endorsements. Some may be known-good-values, but others may be things that are passed on the relying party.

draft-ietf-rats-architecture.md
laurencelundblade: I haven't seen the terms Reference and Invariant Claim before. They seem OK, but they seem synonymous with k-g-v or Reference Values. Are they?

Note also that I see some of these as being passed through the Verifier to the Relying Party.  I described this elaborately here: https://mailarchive.ietf.org/arch/msg/rats/VS-pAO2Io33EagbtpLutIrnUpAQ/

draft-ietf-rats-architecture.md
laurencelundblade: Seemed like the term Endorser was eliminated. I was trying to work around that.

draft-ietf-rats-architecture.md
laurencelundblade: I used the term "secure way" to cover signing and even an in person delivery of flash drive. (I have actually done this with VeriSign root).

draft-ietf-rats-architecture.md
laurencelundblade: Was going to review (and probably comment) on Trust Model next. :-)

draft-ietf-rats-architecture.md
laurencelundblade: I didn't write it. It's what was in the section before. Shows up in the diff because I split it into a separate paragraph. I'm a bit unsure about it myself.

draft-ietf-rats-architecture.md
mcr: I prefer the old text.

laurencelundblade:
Six paragraphs for six characteristics of Endorsements

1) They are absolutely required and the sole basis of a Verifier trusting an Attester and Evidence
2) They are how key material for 1) gets to a Verifier (if not an Endorsement, how else?)
3) They are created by an Endorser
4) They must be conveyed securely from Endorser to Verifyer
5) They may contain known-good or reference values
6) They may contain static implicit claims that are passed to RP's via Attestation Results.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,29 @@ or interval when changes in operational status, health, and so forth occur.
 
 ## Endorsements
 
-An Endorsement is a secure statement that some entity (e.g., a manufacturer) vouches for the integrity of the
-device's signing capability.  For example, if the signing capability is in hardware, then
-an Endorsement might be a manufacturer certificate that signs a public key whose corresponding
-private key is only known inside the device's hardware.  Thus, when Evidence and such an Endorsement
+An Endorsement input to a Verifier is always required. An Endorsement is what allows a Verifier to trust an Attester and in turn to trust the Evidence produced by an Attester. Without Endorsements a Verifier cannot function.
+Said another way, an Attester for which there is no Endorsement is not useful since there is no good basis by which a Verifier can trust it.
```
dthaler, line 738: I still have the same objection as before to this text.   (And I don't think new text is required either.)

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,29 @@ or interval when changes in operational status, health, and so forth occur.
...
+One of the Endorsements for an Attester will just about always contain cryptographic key material.
+This key material will correspond to key material that is put in the Attester.
+The key material is the basis for the Verifier trusting the Attester.
+This architecture does not dictate any particular cryptographic protocol for establishing this trust, hence the key material can be symmetric or asymmetric, a single key, multiple keys, key hierarchies, X.509 cert or chains or other.
+
+Endorsements are produced by an Endorser.
+The Endorser is often the manufacturer of the device that contains the Attester, but not necessarily so.
+For example, it could be an OS vendor that supplies Attester key material to the device/Attester manufacturer.
+
+An Endorsement must be conveyed to the Verifier in a secure way so that the Verifier does not accept fake or false Endorsements that would result in trusting Attesters and Evidence that should not be trusted.
```
dthaler, line 749: I still have the same objection to this text as before.

draft-ietf-rats-architecture.md
```diff
@@ -734,10 +734,29 @@ or interval when changes in operational status, health, and so forth occur.
...
+This architecture does not dictate any particular means of securing this conveyance hence it might be a TLS connection, some sort of signed data, physically transferring the data or other.
```
dthaler, line 750: I don't see the point of calling this out for endorsements more than other messages.   (Personally I think it's already sufficiently covered in the doc for messages in general.)

draft-ietf-rats-architecture.md
laurencelundblade: Do you think that Endorsements are not always required or do you agree with me that they are always required but the text here is redundant?

draft-ietf-rats-architecture.md
laurencelundblade: Sorry, but I'm not able to figure out which previous comment applies to which line here.

draft-ietf-rats-architecture.md
dthaler: Not always required.  The relevant part of my earlier comment above is "If a Verifier owner has some means to establish trust in each Attester's hardware individually then it need not use an endorsement."

draft-ietf-rats-architecture.md
dthaler: It was.  "This implies conveyance via a secure protocol, which I disagree is a must. As long as endorsements are signed, how they're conveyed is irrelevant... conveyance via an insecure protocol is fine."

draft-ietf-rats-architecture.md
laurencelundblade: There are many sentences under Trust Model about the RP trusting the Verifier, more than I have here. 

There are also many sentences about the Endorser trusting the Verifier (the flip of what I am trying to convey) in the Trust Model section.

So there's precedent in the doc for text like I'm proposing and as far as I can see little text that is redundant with mine.

I'm OK taking this out if there's consensus. If we take it out maybe we consider taking out other text about other messages?

draft-ietf-rats-architecture.md
laurencelundblade: I've used the term "secure way" to attempt to cover all manner and means to secure the conveyance, not just to refer to channel-securing protocol like TLS and IPsec. I don't mean to say you must use TLS or IPsec. Sending a signed message over plain HTTP is fine and I would call that a "secure way". Hand carrying a USB flash drive from one company to another is also a "secure way".

So maybe we agree about the concept, but the words need improving?

laurencelundblade: Maybe, text like this: if the Endorsement is not self-securing by signing or such, then some other secure means of conveyance should be used.

laurencelundblade: Added a few sentences to clarify use of signed and unsigned Endorsements.

draft-ietf-rats-architecture.md
laurencelundblade: The trusting of an Attester by a Verifier is to me one of the most fundamental things about attestation and I think it needs to be address prominently and clearly in the architecture document.

I don't know of anyway practical way to establish this trust other than with crypto which of course means some key material in the Attester and related key material in the Verifier.

If an Endorsement is not the one true way the Verifier gets the key it needs, then the architecture should indicate an input to the Verifier by which it does get the key.

I do take it that you agree that these are all valid ways for the Verifier to get the key
 - In an X.509 certificate
 - A CMS-encrypted flat file of keys
 - A flash drive from the Attester maker walked into the server room containing the Verifier
 - TLS-secured access to a database of millions of keys at the Attester manufacturer
 - Many more...

Another option here is to abandon the term Endorsement. It does seem to have some problematic preconception (I keep seem to be guessing wrong, probably because I don't go to the right TCG meetings). Instead of Endorsement these are the verifier inputs:
  - Keys
  - Known-good / Reference values
  - Implicit claims
  - Policy configuration, scripts and such.

dthaler:
Discussion August 25: might be worth mentioning Endorsements in Privacy Considerations section, if Endorsement might contain PII in some use case.

draft-ietf-rats-architecture.md
nedmsmith: r6 contains "Endorser:" definition.

draft-ietf-rats-architecture.md
nedmsmith: 779 should read: "Thus, when Evidence and Endorsements are used together..."

draft-ietf-rats-architecture.md
dthaler: > The trusting of an Attester by a Verifier is to me one of the most fundamental things about attestation and I think it 
> needs to be address prominently and clearly in the architecture document.

Of course.

> I don't know of anyway practical way to establish this trust other than with crypto which of course means some key material in the Attester and related key material in the Verifier.

Yes.

> If an Endorsement is not the one true way the Verifier gets the key it needs, then the architecture should indicate an input to the Verifier by which it does get the key.

If you have an Endorsement, the Verifier needs a way to get the manufacturer's key and establish trust in it.
Without an Endorsement, that same mechanism applies to the Attester's key directly.
Endorsements just provide scalability since you can put one key in your trust anchor store rather than (say) 1 per Attester.

An Endorsement is simply one external key (typically a manufacturer but need not be) vouching for the Attesting Environment's key.

> I do take it that you agree that these are all valid ways for the Verifier to get the key
[...]

Yes

> Another option here is to abandon the term Endorsement.

I would disagree with that option.  The notion of an external key vouching for the Attesting Environment's key is a key concept.   As you state, it's not the only input to the Verifier but that doesn't mean the term isn't an important concept to explain.

PR #105, opened 2020-06-16
working on revising Evidence to close #98

### LINKED ISSUE #98: Definition of "Evidence"
~~~
Based on https://mailarchive.ietf.org/arch/msg/rats/2rwmcm7pQICXN8lzlX21lEkb9Is/
~~~

> Evidence:  A set of information about an Attester that is to be appraised by a Verifier

Kathleen's comment: For evidence, should this definition be more about what evidence is rather than about the roles involved? The attester may not be the same component or system that attests to the evidence as well. Evidence could be measurement data, telemetry (e.g. counters on a system interface - YANG or SNMP), results of a policy comparison to configuration data).

Perhaps:
> Evidence: Information about or originating from a system/firmware/software that may include configuration data, measurements, telemetry, or inferences.

mcr:
This is based upon comments in #98.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -147,7 +147,8 @@ Endorser:
 
 Evidence:
 
-: A set of information about an Attester that is to be appraised by a Verifier
+: A set of information about an Attester that is to be appraised by a Verifier.
+Evidence may include configuration data, measurements, telemetry, or inferences.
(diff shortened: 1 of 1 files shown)


dthaler:
Ok with me, but I thought there was some resistance to "telemetry" so I thought others might want that word removed.  I'm ok either way.

PR #106, opened 2020-06-23
text changes to Hardware Watchdog case, close #102

### LINKED ISSUE #102: Proposed change to section 3.6
~~~
Based on https://mailarchive.ietf.org/arch/msg/rats/2rwmcm7pQICXN8lzlX21lEkb9Is/
~~~

OLD:
> One significant problem is malware that holds a device hostage and does not allow it to reboot to prevent updates to be applied.

NEW:
> One significant problem is malware that holds a device hostage and does not allow it to reboot to prevent updates from being applied.  Remote attestation aids in the verification process that may occur locally as well as it is less likely that the policy or verification processes have also been compromised if hosted on a separate system.

mcr:
text proposal for changes in #102

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -289,7 +289,6 @@ Relying Party:
 ## Hardware Watchdog
 
 One significant problem is malware that holds a device hostage and does not allow it to reboot to prevent updates from being applied.
-Remote attestation aids in the verification process that may occur locally as well as it is less likely that the policy or verification processes have also been compromised if hosted on a separate system.
 This is a significant problem, because it allows a fleet of devices to be held hostage for ransom.
 
 A hardware watchdog can be implemented by forcing a reboot unless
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -286,10 +286,9 @@ Relying Party:
 
 ## Hardware Watchdog
 
-One significant problem is malware that holds a device
-hostage and does not allow it to reboot to prevent updates to be
-applied.  This is a significant problem, because it allows a fleet
-of devices to be held hostage for ransom.
+One significant problem is malware that holds a device hostage and does not allow it to reboot to prevent updates from being applied.
+Remote attestation aids in the verification process that may occur locally as well as it is less likely that the policy or verification processes have also been compromised if hosted on a separate system.
```
mcr, line 290:
```suggestion
```

draft-ietf-rats-architecture.md
```diff
@@ -286,10 +286,9 @@ Relying Party:
...
+This is a significant problem, because it allows a fleet of devices to be held hostage for ransom.
```
mcr, line 291: But, in the case, the Relying Party is the Watch Dog timer in the TPM/secure enclave itself, {{TCGarch}} section 43.3.
So the Attestation Results are returned to the PC, and provided to the enclave.

If the watch dog does not receive regular, and fresh, Attestation Results as to the systems' health, then it forces a reboot.

PR #110, opened 2020-06-26
Verifier can act as Attester to establish trust with Relying Party

gkostal:
Adding a sentence to explicitly call out that a Verifier can act as an attester to establish trust with a Relying Party.


dthaler:
I believe this PR is obsoleted by PR #108

gkostal:
> I believe this PR is obsoleted by PR #108

Yes, I agree.

PR #118, opened 2020-06-30
KM review (part 2) -- Editorial pass on Section 4.4

thomas-fossati:

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -467,8 +467,7 @@ Device, each slot is an Attester, and the main slot is the lead Attester.
 Another example is a multi-chassis router composed of multiple single carrier-grade routers.
 The multi-chassis router provides higher throughput by interconnecting
 multiple routers and can be logically treated as one router for simpler management.
-A multi-chassis router provides a management point and is the only one that
-connects to the Verifier.
+A multi-chassis router provides a management point that connects to the Verifier.
 Other routers are only connected to the main router by the network cables,
 and therefore they are managed and appraised via this main router's help.
 So, in this case, the multi-chassis router is the Composite Device,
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -467,7 +467,8 @@ Device, each slot is an Attester, and the main slot is the lead Attester.
 Another example is a multi-chassis router composed of multiple single carrier-grade routers.
 The multi-chassis router provides higher throughput by interconnecting
 multiple routers and can be logically treated as one router for simpler management.
-Among these routers, there is only one main router that connects to the Verifier.
+A multi-chassis router provides a management point and is the only one that
```
dthaler, line 470: My reading of the old text is that the new text is incorrect.  The multi-chassis router internally includes multiple routers and one of the internal one is referenced, not the overall multi-chassis router.  Is my reading correct?

draft-ietf-rats-architecture.md
```diff
@@ -467,7 +467,8 @@ Device, each slot is an Attester, and the main slot is the lead Attester.
...
+connects to the Verifier.
```
mcr, line 471:
```suggestion
A multi-chassis router provides a management point and is the one that connects to the Verifier.
```

mcr:
@Eliot Lear.

draft-ietf-rats-architecture.md
elear: I will check within Cisco on this.

William-PanWei: The multi-router is a logical concept, it internally includes multiple routers, and these routers are not just the redundancy relationship. From the perspective of the physical routers, there is one router connecting to the Verifier. From the perspective of the logical multi-router, you may say that there is a management point connecting to the Verifier, and actually this point is one router.

PR #119, opened 2020-06-30
KM review (part 2) -- Editorial pass on Section 5

thomas-fossati:

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -529,7 +529,7 @@ which then compares the Attestation Result against its own Appraisal Policy.
 There are three ways in which the process may fail.  First, the Verifier may
 refuse to issue the Attestation Result due to some error in processing, or
 some missing input to the Verifier.
-The second way in which the process may fail is when the Result is examined by
+The second way in which the process may fail is when the Attestation Result is examined by
 the Relying Party, and based upon the Appraisal Policy, the result does not
 pass the policy.
 The third way is when the Verifier is unreachable.
@@ -1283,5 +1283,4 @@ use beyond the period for which it deems the Attestation Result to remain
 valid.  Thus, if the Attestation Result sends a validity lifetime
 in terms of time(RX)-time(RG), then the Relying Party can check
 `time(OP) - time(ER) < time(RX)-time(RG)`.
-
 
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -529,9 +529,9 @@ which then compares the Attestation Result against its own Appraisal Policy.
 There are three ways in which the process may fail.  First, the Verifier may
 refuse to issue the Attestation Result due to some error in processing, or
 some missing input to the Verifier.
-The second way in which the process may fail is when the Attestation Result is examined by
-the Relying Party, and based upon the Appraisal Policy, the result does not
-pass the policy.
+The second way in which the process may fail is when the Attestation Result is 
+examined by the Relying Party, and based upon the Appraisal Policy, the 
+result does not pass the policy.
 The third way is when the Verifier is unreachable.
 
 Since the resource access protocol between the Attester and Relying Party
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -518,9 +529,9 @@ which then compares the Attestation Result against its own Appraisal Policy.
 There are three ways in which the process may fail.  First, the Verifier may
 refuse to issue the Attestation Result due to some error in processing, or
 some missing input to the Verifier.
-The second way in which the process may fail is when the resulting Attestation Result is
-examined by the Relying Party, and based upon the Appraisal Policy, the
-result does not pass the policy.
+The second way in which the process may fail is when the Result is examined by
```
mcr, line 532:
```suggestion
The second way in which the process may fail is when the Attestation Result is examined by
```

draft-ietf-rats-architecture.md
```diff
@@ -518,9 +529,9 @@ which then compares the Attestation Result against its own Appraisal Policy.
 There are three ways in which the process may fail.  First, the Verifier may
 refuse to issue the Attestation Result due to some error in processing, or
 some missing input to the Verifier.
-The second way in which the process may fail is when the resulting Attestation Result is
-examined by the Relying Party, and based upon the Appraisal Policy, the
-result does not pass the policy.
+The second way in which the process may fail is when the Attestation Result is examined by
+the Relying Party, and based upon the Appraisal Policy, the result does not
+pass the policy.
```
mcr, line 534:
```suggestion
The second way in which the process may fail is when the Attestation Result is 
examined by the Relying Party, and based upon the Appraisal Policy, the 
result does not pass the policy.
```

draft-ietf-rats-architecture.md
```diff
@@ -600,24 +609,18 @@ if the Relying Party is a constrained node.
...
-Thus, in this analogy, a prospective volunteer is an Attester, the organization is the Relying Party,
-and a former employer or government agency that issues a report is a Verifier.
-
 ## Combinations
 
-One variation of the background-check model is where the Relying Party
-and the Verifier on the same machine, and so there is no need for a protocol between the two.
+One variation of the background-check model is where the Relying Party and the
+Verifier are on the same machine, performing both functions together.  In this
+case, there is no need for a protocol between the two.
```
mcr, line 616:
```suggestion
One variation of the background-check model is where the Relying Party
and the Verifier are on the same machine, performing both functions together.  
In this case, there is no need for a protocol between the two.
```

draft-ietf-rats-architecture.md
```diff
@@ -600,24 +609,18 @@ if the Relying Party is a constrained node.
...
-and the same device may need to create Evidence for different Relying Parties and different use cases
-(e.g., a network infrastructure device to gain access to the network, and then a
-server holding confidential data to get access to that data).  As such, both models may
-simultaneously be in use by the same device.
+It is also worth pointing out that the choice of model is generally up to the
+Relying Party.  The same device may need to create Evidence for different
+Relying Parties and/or different use cases.  For instance, it would provide
+Evidence to a network infrastructure device to gain access to the network, and
+to a server holding confidential data to gain access to that data. As such,
+both models may simultaneously be in use by the same device.
```
mcr, line 623:
```suggestion
It is also worth pointing out that the choice of model is generally up to the Relying Party.  
The same device may need to create Evidence for different Relying Parties and/or different use cases.  
For instance, it would provide Evidence to a network infrastructure device to gain access to the network, and
to a server holding confidential data to gain access to that data. 
As such, both models may simultaneously be in use by the same device.
```

draft-ietf-rats-architecture.md
```diff
@@ -600,24 +609,17 @@ if the Relying Party is a constrained node.
...
-It is also worth pointing out that the choice of model is generally up to the Relying Party,
-and the same device may need to create Evidence for different Relying Parties and different use cases
-(e.g., a network infrastructure device to gain access to the network, and then a
-server holding confidential data to get access to that data).  As such, both models may
-simultaneously be in use by the same device.
+It is also worth pointing out that the choice of model is generally up to the Relying Party.  
+The same device may need to create Evidence for different Relying Parties and/or different use cases.  
+For instance, it would provide Evidence to a network infrastructure device to gain access to the network, and
+to a server holding confidential data to gain access to that data. As such, 
+both models may simultaneously be in use by the same device.
```
mcr, line 622:
```suggestion
to a server holding confidential data to gain access to that data. 
As such, both models may simultaneously be in use by the same device.
```

PR #121, opened 2020-06-30
KM review (part 2) -- Editorial pass on Section 7.4

thomas-fossati:

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -741,14 +741,14 @@ policy is written in a way that configures the Verifier to trust) a
 manufacturer, or the manufacturer's hardware, so as to be able to
 appraise the trustworthiness of that manufacturer's devices.  In solutions
 with weaker security, a Verifier might be configured to implicitly
-trust firmware or even software (e.g., a hypervisor).  That is, it
-might appraise the trustworthiness of an application component, or operating
-system component or service, under the assumption that information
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -741,14 +741,14 @@ policy is written in a way that configures the Verifier to trust) a
...
-physically resistant to hardware tampering.  The component that is
-implicitly trusted is often referred to as a Root of Trust.
+trust firmware or even software (e.g., a hypervisor).  That is, it might
+appraise the trustworthiness of an application component, operating system
+component, or service under the assumption that information provided about it
+by the lower-layer hypervisor or firmware is true.
+A stronger level of assurance of security comes when information can be vouched
+for by hardware or by ROM code, especially if such hardware is physically
+resistant to hardware tampering.  The component that is implicitly trusted is
+often referred to as a Root of Trust.
```
mcr, line 751:
```suggestion
trust firmware or even software (e.g., a hypervisor).  That is, it
might appraise the trustworthiness of an application component, operating
system component, or service under the assumption that information
provided about it by the lower-layer hypervisor or firmware is true.
A stronger level of assurance of security comes when information can be vouched
for by hardware or by ROM code, especially if such hardware is
physically resistant to hardware tampering.  The component that is
implicitly trusted is often referred to as a Root of Trust.
```

Issue #111, opened 2020-06-30
Appendix A: Time Considerations regressions


dthaler:
PR #88 and PR #91 merged text that I had, and continue to have, problems with.

A summary of the problems follows:
1) It includes events (CC, RP) that are not used in any checks in the text.  I had pointed this out before when the PR was generated, but the PR was merged without addressing this comment.
2) The section about using timestamps with synchronized clocks was removed and replaced with a section that talks about a different mechanism (handles) that is not in any RFC or WG document.   I believe that such replacement is inappropriate.  If the WG adopts a document that talks about a handle distribution mechanism, then a section can be added.  Or if there is a standard from another org that can be referenced then it's fine to add one in that case too. But I think it should not replace the section discussing the well known technique of synchronized clocks (e.g., using secure NTP, or PTP (IEEE 1588-2002) or whatever other standard mechanism).
3) time(HD) is problematic as phrased since all the other events are times as seen by a single entity, whereas time(HD) does not clearly say whose perspective (i.e., whose clock) it is from... is it the distributor's?  Text is unclear.
4) It talks about nonces in a timestamp section, which I think causes confusion since there's already a separately section about nonces and this section is labeled to be scoped to timestamps.   Addressing #2 might address this naturally, if it uses a different section title.
5) There are some editorial typos, such as capital "The" in the middle of a sentence, etc.
6) I have no idea what "delta(time(HD),time(EG))distribution-interval" is supposed to mean.   Normally I would expect an equality for any condition to be evaluated, but I have no idea what this check is.
7) time(HD) and time(EG) are according to different clocks (if my guess about point 3 above is right) and so the delta is not directly comparable.  I'm not sure how point 6 accounts for this.
8) The phrase "A nonce not predictable to an Attester (recentness & uniqueness) is sent to an Attester" has "(recentness & uniqueness)" which doesn't read well (sounds like bad grammar) to me, and is not explained.

henkbirkholz:
Hopefully PR #123 now addresses this issue.

dthaler:
PR #125 reverts the text, leaving the work in progress in PR #111

mcr:
fixed by #145.

PR #123, opened 2020-07-03
time sequence diagram changes (was issue #111) . 

henkbirkholz:
Issue #111 points out that PR #88 and PR #91 have been merged prematurely.

This PR provides a fix for Issue #111:

* the lost "example 3" in Appendix A is re-added
* a new separate 4 is added (the previous example 4 is now example 5)
* the individual detailed comments itemized in Issue #111 should be addressed

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1058,7 +1058,7 @@ or might be defined relative to some other timestamp or timeticks counter.
 | VG | Value generation            | A value to appear in a Claim was created.
 | AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
 | CC | Claim Collection            | An Attesting Environment collects a new/changed Claim value to appear in Evidence.
-| HD | Handle distributed         | A centrally generated handle for time-bound freshness is successfully received by all entities in a domain.
+| HD | Handle distributed          | The (local) time when a centrally generated handle for time-bound freshness is successfully received by each entity in a domain.
 | NS | Nonce sent                  | A nonce not predictable to an Attester (recentness & uniqueness) is sent to an Attester.
 | NR | Nonce relayed               | A nonce is relayed to an Attester by another entity.
 | EG | Evidence generation         | An Attester creates Evidence from collected Claims (CC).
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1253,6 +1253,7 @@ In this example, the handle is composed of a signed timestamp. If the Attester d
 The validity span of a handle can be encoded into the handle itself or determined by an Appraisal Policy.
 From the point of view of a Verifier, the generation of valid Evidence is only possible, if the age of the handle used in the Evidence generation is younger than the duration of the distribution interval -- "delta(time(HD),time(EG)) \< delta(time(HD),time(HD'))".
 
+[XXX- this requires an adopted cite-able document]
 ## Example 5: Nonce-based Background-Check Model Example
 
 The following example illustrates a hypothetical Background-Check Model
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1224,23 +1224,32 @@ This eliminates the need for clocks between the entities depicted in the diagram
 Additionally, latency jitter for time(HD) between all entities involved (introduced by the conveyance of handles) is assumed to be the less than the time drift of a local clock with common precision.
 
 ~~~~
-.----------.         .---------------.              .----------.
-| Attester |         | Relying Party |              | Verifier |
-'----------'         '---------------'              '----------'
-  time(VG)                   |                           |
-        |                    |                           |
-     ---+----time(HD)--------+------time(HD))------------+---
-        |                    |                           |
-  time(AA)                   |                           |
-        |                    |                           |
-  time(EG)                   |                           |
-        |----Evidence------->|                           |
-        |    {time(EG)}   time(ER)--Evidence{time(EG)}-->|
-        |                    |                        time(RG)
-        |                 time(RA)<-Attestation Result---|
-        |                    |        {time(RX)}         |
-        ~                    ~                           ~
-        |                    |                           |
+                                                               .------------.
+                                                               | handle     |
+                                                               | distributor|
+                                                               '------------'
+.----------.         .---------------.              .----------.    | | |
+| Attester |         | Relying Party |              | Verifier |    | | |
+'----------'         '---------------'              '----------'    | | |
+  time(VG)                   |                           |          | | |
+        |                    |                           |          | | |
+        |                    |<-------handle------------------------/ | |
+        |                 time(HD)                       |            | |
+        |<-------handle-----------------------------------------------/ |
+  time(HD)                   |                           |<---handle----/
+        |                    |                        time(HD)          .
+  time(AA)                   |                           |              .
+        |                    |                           |              .
+  time(EG)                   |                           |   {}-on-wire .
+        |----Evidence------->|                           |              .
+        |    {time(EG)}   time(ER)--Evidence{time(EG)}-->|              .
+        |                    |                        time(RG)          .
+        |                 time(RA)<-Attestation Result---|              .
+        |                    |        {time(RX)}         |              .
+        ~                    ~                           ~            ...
+        |                 time(HD')<----------------------------------/..
+   time(HD')<----------------------------------------------------------/.
+        |                    |                        time(HD')<--------/
         |                 time(OP)                       |
 ~~~~
 
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1058,7 +1058,7 @@ or might be defined relative to some other timestamp or timeticks counter.
 | VG | Value generation            | A value to appear in a Claim was created.
 | AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
 | CC | Claim Collection            | An Attesting Environment collects a new/changed Claim value to appear in Evidence.
-| HD | Handle distribution         | A centrally generated identifier for time-bound recentness across a domain of devices is successfully distributed to Attesters.
+| HD | Handle distributed         | A centrally generated handle for time-bound freshness is successfully received by all entities in a domain.
 | NS | Nonce sent                  | A nonce not predictable to an Attester (recentness & uniqueness) is sent to an Attester.
 | NR | Nonce relayed               | A nonce is relayed to an Attester by another entity.
 | EG | Evidence generation         | An Attester creates Evidence from collected Claims (CC).
@@ -1216,9 +1216,12 @@ The time considerations in this example are equivalent to those discussed under 
 ## Example 4: Handle-based Background-Check Model Example
 
 The following example illustrates a hypothetical Background-Check Model
-solution that uses centrally generated identifiers for explicit time-keeping (referred to as "handle" in this example).
-Handles can be qualifying data, such as nonces or signed timestamps. In this example, centrally generated signed timestamps and -- and synchronized clocks between all entities -- are distributed
-in periodic intervals as handles.  If the Attester lacks a source of time based on an absolute timescale, a relative source of time, such as a tick counter can be used, alternatively.  In this example, evidence generation is not triggered at value generation, but at events at which the Attesting Environment becomes of changes to the Target Environment.
+solution that uses centrally generated handles for explicit time-keeping (using local clocks) or implicit time-keeping (using trigger events such as receiving a handle or becoming aware of a changed Claim value).
+Handles are be qualifying data, such as nonces or signed timestamps.
+In this example, centrally generated signed timestamps are used as handles and are distributed in periodic intervals at time(HD).
+The distributor of handles includes a trustable source of time to produce the signed timestamps.
+This eliminates the need for clocks between the entities depicted in the diagram below. If entities include at least a relative clock, such as a tick counter, they can roughly determine (due to potentially latency during handle distribution) when the handles cannot be considered fresh anymore.
+Additionally, latency jitter for time(HD) between all entities involved (introduced by the conveyance of handles) is assumed to be the less than the time drift of a local clock with common precision.
 
 ~~~~
 .----------.         .---------------.              .----------.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -1208,38 +1236,16 @@ in periodic intervals as handles.  If the Attester lacks a source of time based
...
+This on-change generated Evidence can be based on a mix of relevant "old" Claim values and just recently generated and collected "new" Claim values.
+The delta between time(VG) and time(AA) -- "delta(time(VG),time(AA))" -- can be very large for corresponding "old" Claim values.
+
+In order to be able to generate valid on-change Evidence, the Attester requires a fresh (i.e. not expired) handle that has been distributed to all involved entities.
+In this example, the handle is composed of a signed timestamp. If the Attester does not include a clock, it is possible that the Attester uses an expired handle and it is the Verifier's duty to detect that during the appraisal of Evidence.
+The validity span of a handle can be encoded into the handle itself or determined by an Appraisal Policy.
+From the point of view of a Verifier, the generation of valid Evidence is only possible, if the age of the handle used in the Evidence generation is younger than the duration of the distribution interval -- "delta(time(HD),time(EG)) \< delta(time(HD),time(HD'))".
+
```
mcr, line 1247:
```suggestion

[XXX- this requires an adopted citeable document]
```

draft-ietf-rats-architecture.md
```diff
@@ -1052,7 +1052,7 @@ or might be defined relative to some other timestamp or timeticks counter.
 | VG | Value generation            | A value to appear in a Claim was created.
 | AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
 | CC | Claim Collection            | An Attesting Environment collects a new/changed Claim value to appear in Evidence.
-| HD | Handle distribution         | A centrally generated identifier for time-bound recentness across a domain of devices is successfully distributed to Attesters.
+| HD | Handle distributed         | A centrally generated handle for time-bound freshness is successfully received by all entities in a domain.
```
mcr, line 1055: Whose clock is addressed by time HD?

draft-ietf-rats-architecture.md
```diff
@@ -1058,7 +1058,7 @@ or might be defined relative to some other timestamp or timeticks counter.
 | VG | Value generation            | A value to appear in a Claim was created.
 | AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
 | CC | Claim Collection            | An Attesting Environment collects a new/changed Claim value to appear in Evidence.
-| HD | Handle distribution         | A centrally generated identifier for time-bound recentness across a domain of devices is successfully distributed to Attesters.
+| HD | Handle distributed          | The (local) time when a centrally generated handle for time-bound freshness is successfully received by each entity in a domain.
```
dthaler, line 1061: the term "domain" is undefined
```suggestion
| HR | Handle received          | A centrally generated handle for time-bound freshness is received.
```
None of the others start with "The time when", it's all about the event not the time.
Also if it's time received, then "distributed" is the wrong term, "received" should be used.

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
 
 ## Example 3: Timestamp-based Background-Check Model Example
 
+The following example illustrates a hypothetical Background-Check Model solution that uses timestamps and requires roughly synchronized clocks between the Attester, Verifier, and Relying Party.
```
dthaler, line 1191: Isn't this part covered by PR 124 now?   can remove from this one or abandon this one if other PRs will cover it

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+
+The time considerations in this example are equivalent to those discussed under Example 1 above.
+
+## Example 4: Handle-based Background-Check Model Example
+
 The following example illustrates a hypothetical Background-Check Model
-solution that uses centrally generated identifiers for explicit time-keeping (referred to as "handle" in this example).
-Handles can be qualifying data, such as nonces or signed timestamps. In this example, centrally generated signed timestamps and -- and synchronized clocks between all entities -- are distributed
-in periodic intervals as handles.  If the Attester lacks a source of time based on an absolute timescale, a relative source of time, such as a tick counter can be used, alternatively.  In this example, evidence generation is not triggered at value generation, but at events at which the Attesting Environment becomes of changes to the Target Environment.
+solution that uses centrally generated handles for explicit time-keeping (using local clocks) or implicit time-keeping (using trigger events such as receiving a handle or becoming aware of a changed Claim value).
```
dthaler, line 1217: I can't understand the phrase from "for explicit" until the end of the sentence.

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+Handles are be qualifying data, such as nonces or signed timestamps.
```
dthaler, line 1218: Qualifying for what?  And we already have other sections for nonces and timestamps, I don't follow what's different about this one (especially why it would combine nonce & timestamp into one example when they're separated in the other examples)

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+In this example, centrally generated signed timestamps are used as handles and are distributed in periodic intervals at time(HD).
```
dthaler, line 1219: As above, I think "distributed" is the wrong word.   And the "periodic" might(?) be too strong if talking about receive time since there might be significant jitter due to congestion, so it may not look very periodic?

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+The distributor of handles includes a trustable source of time to produce the signed timestamps.
+This eliminates the need for clocks between the entities depicted in the diagram below. If entities include at least a relative clock, such as a tick counter, they can roughly determine (due to potentially latency during handle distribution) when the handles cannot be considered fresh anymore.
```
dthaler, line 1221: "potentially latency" is bad grammar.  "potential latency"?

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+                                                               '------------'
+.----------.         .---------------.              .----------.    | | |
+| Attester |         | Relying Party |              | Verifier |    | | |
+'----------'         '---------------'              '----------'    | | |
+  time(VG)                   |                           |          | | |
+        |                    |                           |          | | |
+        |                    |<-------handle------------------------/ | |
+        |                 time(HD)                       |            | |
+        |<-------handle-----------------------------------------------/ |
+  time(HD)                   |                           |<---handle----/
```
dthaler, line 1237: this diagram is confusing since there are three time(HD) that all mean different things.
Suggest using ' and '' (prime and double prime) or something similar

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
-  time(EG)                   |                           |
-        |----Evidence------->|                           |
-        |    {time(EG)}   time(ER)--Evidence{time(EG)}-->|
-        |                    |                        time(RG)
-        |                 time(RA)<-Attestation Result---|
-        |                    |        {time(RX)}         |
-        ~                    ~                           ~
-        |                    |                           |
+                                                               .------------.
+                                                               | handle     |
```
dthaler, line 1226:
```suggestion
                                                               | Handle     |
```

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
-heart-beat timeout). From the point of view of a Verifier, validity of
-Evidence is only given if the handle used in Evidence satisfies
-delta(time(HD),time(EG))distribution-interval.
-
-In this usage scenario, time(VG), time(AA), and time(EG) are tightly
-coupled. Also, the absolute point in time at which a handle is received
-by all three entities is assumed to be close to identical.
-
-## Example 4: Nonce-based Background-Check Model Example
+At time(AA), the Attesting Environment is able to trigger an event (e.g. based on an Event-Condition-Action model) to generate Evidence as soon/recent as possible at time (EG).
```
dthaler, line 1254: Why do you need time(AA) and time(EG) both, instead of just using one, like time(EG) as in other examples?  In particular, there is no security check mentioned below that uses time(AA) and so that event should not be added

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+This on-change generated Evidence can be based on a mix of relevant "old" Claim values and just recently generated and collected "new" Claim values.
```
dthaler, line 1255: How is this sentence different from the other examples? (If it's not, then remove the sentence.)

draft-ietf-rats-architecture.md
```diff
@@ -1188,64 +1188,80 @@ in terms of time(RX)-time(RG), then the Relying Party can check
...
+The delta between time(VG) and time(AA) -- "delta(time(VG),time(AA))" -- can be very large for corresponding "old" Claim values.
+
+In order to be able to generate valid on-change Evidence, the Attester requires a fresh (i.e. not expired) handle that has been distributed to all involved entities.
+In this example, the handle is composed of a signed timestamp. If the Attester does not include a clock, it is possible that the Attester uses an expired handle and it is the Verifier's duty to detect that during the appraisal of Evidence.
+The validity span of a handle can be encoded into the handle itself or determined by an Appraisal Policy.
+From the point of view of a Verifier, the generation of valid Evidence is only possible, if the age of the handle used in the Evidence generation is younger than the duration of the distribution interval -- "delta(time(HD),time(EG)) \< delta(time(HD),time(HD'))".
```
dthaler, line 1261: There are three different time(HD)'s in the diagram, which one is this talking about?
Same for time(HD').
The sentence seems to imply it's talking about the Verifiers, but if it's just the Verifier then why are there lines to the Attester and Relying Party, with time(HD) values not used in any security check or protocol message.
This doesn't make sense.

draft-ietf-rats-architecture.md
henkbirkholz: The relevant part here is the distribution of handles. This allows for implicit time-keeping only: knowledge about "yes I have knowledge of that handle or not" is the only relevant part here. Therefore, the focus here is on very implicit time measures. Naturally they are basically on the other end of the spectrum that are explicit timestamps. "Awareness of changes" & "Reception of Handle" are the relevant parts here. It is also important that the nature of the Handle itself, be it Nonces or Timestamps, or other concepts is explicitly allowed and encouraged. That is the essence of the example: Triggers based on implicit time-keeping, such as awareness of changes -> the reception of handles is one of them, the changes of values another.

=> Trustable freshness with the least amount of clocks used (but you still can use clocks!)

In consequence: I approve your change. In each iteration we try to align your input with the concept itself.

draft-ietf-rats-architecture.md
henkbirkholz: That is probably an artifact.

draft-ietf-rats-architecture.md
henkbirkholz: We were trying not to "make up terms" in the text for readability. From the top if my head, I can't remember why we abandoned the definition of explicit and implicit time-keeping in this section. But we discussed it. Would it help to introduce the terms earlier in the text? If not, what would help you in understanding this sentence better (except reading it until the end)? I am shying away a bit from adding more words to fix other words right now.

draft-ietf-rats-architecture.md
henkbirkholz: Things that are typically qualified are: origin of handle or intend of handle. E.g. if your handle is the hash of an handle that is an external timestamp + local clock time + proof of synchronization of the two => you get TUDA. If your handle is the hash of a received nonce received during subscripton, a relative tick-counter originating from your Attesting Environment, and a reference to a global time-scale => you get Attestation Event Stream Subscription.

In general, qualifying data is used in the creation of evidence. It is a annotation of intend and context to a value. The new Restful Resources is showing a good subset in which very explicit rules enrich nonces so they become qualified data. The construction rules defined there provide a whole new spectrum of meanings/semantics to nonces (which TUDA also does, as does network device description, please see below). It is important to highlight that this "semantic enrichment" is allowed, actually used quite often in the field, and goes beyond what is typically perceived as the use of nonces (we also struggle with that until a few weeks ago, when there still was a ver lob-sided focus on reply prevention in the freshness section). It is quite useful to dedicate at least one example in an Appendix to this topic, I think.

Personally, i thought "qualifying data" is a well known term, such as qualified name or such. So my ask is: is this a general misconception of mine? Think of it as nonces with a bigger scope semantics (which we call handles here). If qualifying data is really not self-explanatory, what would be a better phrasing?

draft-ietf-rats-architecture.md
henkbirkholz: I am okay with using received instead of distributed here, if that helps.

We tried to simplify text here again, to make it more readable. With simplification you sometimes break meaning, unfortunately. I admit that the resulting abstraction (boiling it down to simply periodic, which is not wrong per se, but is imprecise) is a compromise. What do others think?

draft-ietf-rats-architecture.md
henkbirkholz: We tried to pick up on the notion that time(HD) is not exactly the same time, but "good enough" or "close enough" to the same time. If your time resolution is fine grained enough. The highlighted three time(HD) are different of course.

Again, the proposal to use three lines instead of one was made at the last call. If seeing this as ascii art makes us not using three lines in the end, but one. I am fine with that. We still have to highlight in text that there is some variance due to network conveyance here. This is how things work today - we simply need a way to depict that in an example here. I am open to suggestions and would be okay with "one line".

draft-ietf-rats-architecture.md
henkbirkholz: Because time(AA) does not seem to be relevant wrt how solutions work in the other examples. Yet it is the vital point in time that drives this example (potentially no clocks involved, a lot of types of handles allowed). We wrote text version multiple times until now and it results in a strange disconnect when you are very explicit about time(foo) and then suddenly you just use English text to express: "and then the most important time is when the Attesting Environment becomes aware of changes to some relevant values in Target Environments, starts to collect them in order to create the semantic Claims and finally generates the Evidence at time(EG)". There is a very confusing disconnect for the reader, why only the last time is expressed so explicitly, while the potential latency between VG and AA is the driver (or disabler!) for secure evidence here.

We are now maybe too reluctant to add more complexity and text right now. Maybe that is the primary cause for confusion, still. The security check you expect here is other an implementation specific one (is that okay to highlight here in the architecture) or not done by the Attester but the Verifier or even at the Attestation Result appraisal stage at the Relying Party. That would create a lot of text to spell all that out. Maybe there is a simpler way to phrase it? When we read this in the editor's session, we should discuss this live.

draft-ietf-rats-architecture.md
henkbirkholz: In practice today, it is not possible to include time(VG) as a value into evidence. Sometimes, the truth is that you can only derive time(AA). And time(VG) != time(AA). It gets even "worse" when you have no local clock and your "measure of time" is only "which handles did I already get?" All these scenarios are solvable and we do not want to include a complete problem statement here, just an example that readers faced with these kinds of challenges today can relate with.

I understand that you would like to subsume that context under time(EG) and I see how that reads simpler. I have to restate that I remain a bit worried, though, that this simplification does not reflect the real world,and could potentially lead to wrong interpretation of the security assumptions/consideration.

draft-ietf-rats-architecture.md
henkbirkholz: I assume this is addressed by one of my earlier replies. This is one of the topics that should be discussed later today (20-07-07).

draft-ietf-rats-architecture.md
henkbirkholz: Handles a created at time(X) and received at time(Y). time(X) is typically easily obtainable and non-ambiguous, but not very useful. time(Y) introduces a variance of uncertainty across a certain time window. Effectively, that keeps them at the same point of time in low resolution timestamps and, analogously, not, if the resolution of timestamps is fine-grained enough. As we successful avoided to talk about timestamp resolution (which is essential to remote attestation, but not to this document), we'd like to keep it that way. Therefore, the choice if it is handle generation, handle emission, or handle reception is up to us here. I'd preferred distribution (encompassing the latency by referencing the start of distribution -> emission), but I can see how the perception is always falling back to "local clocks" even if they might not be there.

Some high level context: in the end, this is for us to decide. Handle Distribution is used a lot in constraint node environments.
Sometimes things have clocks, sometimes they don't.
Sometimes you want to use a trustable central times authority, sometimes you don't.
Sometimes you can infer a time when values were generated, sometimes you can't.
Depicting permutations of combinations of all existing factors would result in 32+ examples. So that did not seem reasonable.

Hence, example 4 is composite of the most typical constraints that you encounter in IoT:

* sometimes no clocks
* sometimes no central time
* sometimes no capability for value monitoring

My personal goal is to capture these three essential items into example 4. Any help is welcome!

PR #124, opened 2020-07-03
restore example 3

mcr:
This splits example 3 out of pull request #123

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1188,6 +1188,31 @@ in terms of time(RX)-time(RG), then the Relying Party can check
 
 ## Example 3: Timestamp-based Background-Check Model Example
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -1188,6 +1188,31 @@ in terms of time(RX)-time(RG), then the Relying Party can check
 
 ## Example 3: Timestamp-based Background-Check Model Example
 
+The following example illustrates a hypothetical Background-Check Model solution that uses timestamps and requires roughly synchronized clocks between the Attester, Verifier, and Relying Party.
```
dthaler, line 1191:
```suggestion
The following example illustrates a hypothetical Background-Check Model
solution that uses timestamps and requires roughly synchronized
clocks between the Attester, Verifier, and Relying Party.
```
(Put line breaks back where they used to be in the source)

dthaler:
Also covered in PR #125

PR #130, opened 2020-07-09
Revice Privacy Considerations

laurencelundblade:


draft-ietf-rats-architecture.md
```diff
@@ -973,28 +973,40 @@ A more detailed discussion with examples appears in {{time-considerations}}.
...
-
-Furthermore, because Evidence might contain sensitive information,
-Attesters are responsible for only sending such Evidence to trusted
-Verifiers.  Some Attesters might want a stronger level of assurance of
-the trustworthiness of a Verifier before sending Evidence to it.  In such cases,
-an Attester can first act as a Relying Party and ask for the Verifier's own
-Attestation Result, and appraising it just as a Relying Party would appraise
-an Attestation Result for any other purpose.
+Every claim in Attestation Evidence and Attestation Results is potentially PII (Personally Identifying Information) depending on the end-end use case of the attestation.
+Each deployment of attestation must take into account how each claim may or may not be used in a privacy violating manner by the relying parties receiving the attestation.
```
henkbirkholz, line 977:
```suggestion
Each deployment of remote attestation must take into account how each Claim may or may not be used in a privacy violating manner by Verifiers receiving Evidence or Relying Parties receiving Attestation Results.
```

draft-ietf-rats-architecture.md
```diff
@@ -973,28 +973,40 @@ A more detailed discussion with examples appears in {{time-considerations}}.
...
-kinds of information are disclosed, and to whom they are exposed.
-
-Furthermore, because Evidence might contain sensitive information,
-Attesters are responsible for only sending such Evidence to trusted
-Verifiers.  Some Attesters might want a stronger level of assurance of
-the trustworthiness of a Verifier before sending Evidence to it.  In such cases,
-an Attester can first act as a Relying Party and ask for the Verifier's own
-Attestation Result, and appraising it just as a Relying Party would appraise
-an Attestation Result for any other purpose.
+Every claim in Attestation Evidence and Attestation Results is potentially PII (Personally Identifying Information) depending on the end-end use case of the attestation.
```
henkbirkholz, line 976:
```suggestion
Every Claim in attestation Evidence and Attestation Results is potentially PII (Personally Identifying Information) depending on the end-end use case of the remote attestation procedure.
```

draft-ietf-rats-architecture.md
```diff
@@ -973,28 +973,40 @@ A more detailed discussion with examples appears in {{time-considerations}}.
...
+Each deployment of attestation must take into account how each claim may or may not be used in a privacy violating manner by the relying parties receiving the attestation.
+
+In cases where the device sending attestation is owned by the relying party, little to no consideration for privacy is needed. In other cases, such as when a mobile phone is used to access a large number of web sites, apps and services, privacy must be considered very thoroughly.
```
henkbirkholz, line 979:
```suggestion
In cases where the Attester sending Evidence is owned by the Relying Party Owner, little to no consideration for privacy is needed. In other cases, such as when a mobile phone is used to access a large number of web sites, apps and services, privacy must be considered very thoroughly.
```

draft-ietf-rats-architecture.md
```diff
@@ -973,28 +973,40 @@ A more detailed discussion with examples appears in {{time-considerations}}.
...
+Evidence and Attestation Results data structures are expected to support integrity protection encoding (e.g., COSE, JOSE, X.509) and optionally might support confidentiality protection (e.g., COSE, JOSE).
+If confidentiality protection is omitted or unavailable, the protocols that convey Evidence or Attestation Results should provide confidentiality.
+
+The Verifier may or may not be trusted with privacy sensitive claims depending on the use case and nature of the Verifier.
+For example, if the Verifier is operated by the Relying Party and that Relying Party should not be allowed to access to privacy violating claims, those claims must not be in Attestation Evidence.
+On the other hand, the Verifier might be designed and operated to be a privacy proxy in which case part of its job is to remove or anonymize privacy violating claims in Attestation Evidence so they are not passed on in Attestation Results.
+
+Thus, in use cases where privacy violating claims are in the Attestation Evidence, the Attester should authenticate the Verifier to know that it is trusted.
```
henkbirkholz, line 1008:
```suggestion
Thus, in usage scenarios where privacy violating Claims are in the attestation Evidence, an Attester should authenticate Verifiers to know that they are trusted. 
```

draft-ietf-rats-architecture.md
henkbirkholz: Not entirely sure that this suggested change is not breaking the intent of the original text.

draft-ietf-rats-architecture.md
dthaler: I don't think it's true that every claim is PII.  If someone adds a claim saying it supports a specific version of the EAT spec, just like every other device, I wouldn't call that PII.  Similarly if there is a claim that contains the current time, or a nonce supplied by a verifier, those wouldn't be PII either under any use case I can think of.   The key question is whether it significantly narrows the anonymity set (as discussed in RFC 6973).

dthaler: I also think the original text in red needs to stay at the top.  PII is not the only aspect of privacy, it's just an important subset.

draft-ietf-rats-architecture.md
```diff
@@ -973,28 +973,40 @@ A more detailed discussion with examples appears in {{time-considerations}}.
...
+These are some strategies that can be used to address the privacy issue:
+
+* Determine that a particular claim is not privacy violating in the limited context of use.
+  For example, a router sending its serial number to the network management center that owns the router violates no one’s privacy.
+
+* Simply omit the claim. For example, omit the serial number from attestations sent by a phone to web sites it connects to.
+
+* Prompt the user. For example, ask the user if it is OK for a particular web site to access location data.
+
+* Use a privacy proxy service between the device and the relying party to remove or anonymize the privacy violating claims.
```
nedmsmith, line 990: change 'device' to attester to be consistent use of roles terminology

draft-ietf-rats-architecture.md
nedmsmith: Strike "Every", change "Claim" to "Claims" and change "is" to "are". Then I'm OK with the proposed new text.

draft-ietf-rats-architecture.md
nedmsmith: The text after "Verifiers" and "Relying Parties" is redundant / unnecessary since we explain in other sections the function of Verifiers and Relying Parties. Strike "receiving Evidence" and "receiving Attestation Results"

draft-ietf-rats-architecture.md
nedmsmith: Maybe avoid using "trusted" and instead say "Verifiers that have acceptable privacy policies".

draft-ietf-rats-architecture.md
nedmsmith: "In cases where Evidence is owned by the Relying Party Owner, consideration for privacy may be of lesser importance. In other cases where Evidence is owned by the Attester, such as when a mobile phone is used to access a large number of web sites, applications or services, consideration for privacy may be of greater importance."

mcr (changes requested):
consider comments and try something else.

PR #143, opened 2020-08-31
Refer to Owner rather than administrator.

laurencelundblade:
Also, my assumption is that all messages, configs and such go over one of the arrows in the figure. There are no back channels for things that are vital to or a major part of attestation. Seems like it would be a poor document if the major parts of attestation are not explicitly carried in the flows depicted.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -360,13 +360,12 @@ The Verifier uses the Evidence, and any Endorsements from Endorsers,
 by applying an Appraisal Policy for Evidence to assess the trustworthiness of the Attester,
 and generates Attestation Results for use by Relying Parties.  The Appraisal Policy for Evidence
 might be obtained from an Endorser along with the Endorsements, or might be obtained
-via some other mechanism such as being configured in the Verifier by an
-administrator.
+via some other mechanism such as being configured in the Verifier by the Verifier Owner.
 
 The Relying Party uses Attestation Results by applying its own
 Appraisal Policy to make application-specific decisions such as authorization decisions.
-The Appraisal Policy for Attestation Results might, for example, be configured in the Relying Party
-by an administrator.
+The Appraisal Policy for Attestation Results is configured in the Relying Party
+by the Relying Party Owner.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -360,13 +360,12 @@ The Verifier uses the Evidence, and any Endorsements from Endorsers,
...
 might be obtained from an Endorser along with the Endorsements, or might be obtained
-via some other mechanism such as being configured in the Verifier by an
-administrator.
+via some other mechanism such as being configured in the Verifier by the Verifier Owner.
 
 The Relying Party uses Attestation Results by applying its own
 Appraisal Policy to make application-specific decisions such as authorization decisions.
-The Appraisal Policy for Attestation Results might, for example, be configured in the Relying Party
-by an administrator.
+The Appraisal Policy for Attestation Results is configured in the Relying Party
```
William-PanWei, line 367: I prefer ‘might be’ here, and configuration may be just one of the possible mechanisms. Suggest changing it back.

draft-ietf-rats-architecture.md
```diff
@@ -360,13 +360,12 @@ The Verifier uses the Evidence, and any Endorsements from Endorsers,
...
+by the Relying Party Owner.
```
mcr, line 368:
```suggestion
The Appraisal Policy for Attestation Results is configured in the Relying Party by the Relying Party Owner, 
and/or is programmed into the Relying Party.
```

draft-ietf-rats-architecture.md
```diff
@@ -360,13 +360,12 @@ The Verifier uses the Evidence, and any Endorsements from Endorsers,
 by applying an Appraisal Policy for Evidence to assess the trustworthiness of the Attester,
 and generates Attestation Results for use by Relying Parties.  The Appraisal Policy for Evidence
 might be obtained from an Endorser along with the Endorsements, or might be obtained
```
mcr, line 362:
```suggestion
might be obtained from an Endorser along with the Endorsements, and/or might be obtained
```

PR #144, opened 2020-09-01
Annotate incomparable timestamps with a clock-specific suffix

dthaler:
Per offline discussion between Henk and Dave, trying this annotation to see if it helps

Also try combining VG and AA into VC since nothing in doc uses AA, and any example could use either VG or AA, so combining makes any example work for both.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1149,7 +1149,7 @@ or might be defined relative to some other timestamp or timeticks counter.
 
 | ID | Event                       | Explanation of event
 |----|-----------------------------|-----------------------
-| VC | Value created               | A value to appear in a Claim was created.  In some cases, a value may have technically existed before an Attester became aware of it but the Attester might have no idea how long it has had that value.  In such a case, the Value created time is the time at which the Claim containing the copy of the value was created.
+| VG | Value generated             | A value to appear in a Claim was created.  In some cases, a value may have technically existed before an Attester became aware of it but the Attester might have no idea how long it has had that value.  In such a case, the Value created time is the time at which the Claim containing the copy of the value was created.
 | HD | Handle distribution         | A centrally generated identifier for time-bound recentness across a domain of devices is successfully distributed to Attesters.
 | NS | Nonce sent                  | A nonce not predictable to an Attester (recentness & uniqueness) is sent to an Attester.
 | NR | Nonce relayed               | A nonce is relayed to an Attester by another entity.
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -1226,57 +1227,64 @@ The following example illustrates a hypothetical Passport Model
...
 that the Attestation Result vouches for the Attester's signing
 capability.
 
 The Relying Party must still be careful, however, to not allow continued
 use beyond the period for which it deems the Attestation Result to remain
 valid.  Thus, if the Attestation Result sends a validity lifetime
-in terms of time(RX)-time(RG), then the Relying Party can check
-`time(OP) - time(NS') < time(RX)-time(RG)`.
+in terms of time(RXv)-time(RGv), then the Relying Party can check
+`time(OPr)-time(NSr) < time(RXv)-time(RGv)`.
```
William-PanWei, line 1287: Here is a little weird to me. Because even when the actual time(NSr) > actual time(RXv), 'time(OPr)-time(NSr) < time(RXv)-time(RGv)' may still be true. If so this comparsion makes no sense.
One better expression here, I think, is that 'time(OPr)-time(NSr) + time(RRa)-time(EGa) < time(RXv)-time(RGv)'.

William-PanWei (changes requested):

draft-ietf-rats-architecture.md
```diff
@@ -1149,8 +1149,7 @@ or might be defined relative to some other timestamp or timeticks counter.
 
 | ID | Event                       | Explanation of event
 |----|-----------------------------|-----------------------
-| VG | Value generation            | A value to appear in a Claim was created.
-| AA | Attester awareness          | An Attesting Environment starts to be aware of a new/changed Claim value.
+| VC | Value created               | A value to appear in a Claim was created.  In some cases, a value may have technically existed before an Attester became aware of it but the Attester might have no idea how long it has had that value.  In such a case, the Value created time is the time at which the Claim containing the copy of the value was created.
```
henkbirkholz, line 1152:
```suggestion
| VG | Value generated             | A value to appear in a Claim was created.  In some cases, a value may have technically existed before an Attester became aware of it but the Attester might have no idea how long it has had that value.  In such a case, the Value created time is the time at which the Claim containing the copy of the value was created.
```

henkbirkholz: Less growing pains, more consistence?

henkbirkholz: The underlying question here is. Is it necessary that some Verifiers are informed about the "timeliness" of the "new VG"?

As Dave already highlighted, we discussed this a bit and the sanity check ask is the following:

If we collapse VG and AA semantics, it is not distinguishable for Verifier, if a corresponding Value generated time is about Attester awareness/Claim generation or Value Generation. Is that okay or does that create uncertainty that could result in degradation of evidence trustworthiness?

PR #141, opened 2020-08-31
Tighten up terminology definitions.

laurencelundblade:

PR #147, opened 2020-09-08
add reference values to diagram

mcr:
change diagram to add reference values as a distinct input to verifier
This will go with a TBD pull request from @dthaler about where the text goes, in an effort to close up #141


dthaler:
I incorporated this (after trying to pretty it up some more) in PR #150.

PR #148, opened 2020-09-10
System Examples Appendix

laurencelundblade:
Here's a first rough draft of an appendix of end-end system examples that show how Endorsements, Verifier, Passport/Background and such apply (or don't apply) to some examples.   I think this conveys many of the things I've been trying largely unsuccessfully and relieves the main text of trying to explain them. I think this an important addition.

It is still a rough draft to see if there is interest.


draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
 valid.  Thus, if the Attestation Result sends a validity lifetime
 in terms of time(RX)-time(RG), then the Relying Party can check
 `time(OP) - time(ER) < time(RX)-time(RG)`.
+
+# Appendix B: System Examples
```
mcr, line 1321: I think that by "System", you mean something specific, yet the text is rather general.  I suggest making the example even more specific: pick one of Smartphone. Desktop PC. VoIP phone. Smart IoT Speaker.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+(Rough first draft to see if there is interest)
+
+## System Example with OS Vendor as Main Driver
+
+Three entities (companies / organizations) participate in this examples.
+
+The operating system vendor is at the center. It sets the architecture and the level of security achieved. It performs the following:
+* Generates large quantities of ECDSA key pairs to be used as per-device attestation keys
+* Signs the public keys using an attestation PKI and root that they control creating per-device certificates.
+* Securely sends the private keys and certificates to the device OEM in batches to be used in manufacturing
```
mcr, line 1332: Assumes factory generated keys, which is a reasonable choice.  I suggest making this choice explicit.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Authors the device SW that is the Attester and ships it to the OEMs. This is a part of the OS.
```
mcr, line 1333: "This is part of the OS" is probably incorrect for many situations.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Publishes the root certificate and definition of the claims
+
+The device OEM does the manufacturing. It performs the following:
+* Physically makes the device to OS vendors security requirements
+* Installs the OS on the device
+* Personalizes each device with a unique private key and certificate supplied by the OS vendor
+
+The application vendor that implements some application with a back-end cloud service. It performs the following:
+* Author an app on the device that calls the attestation API to obtain Attestation Evidence
```
mcr, line 1342:
```suggestion
* Author software that runs on the device that calls the attestation API to obtain Attestation Evidence
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+Relying Party and Relying Party Owner:
+
+: are also the application vendor cloud service; since the Verify and Relying Part are the same the distinction between passport and background check doesn’t apply.
+
+Attestation Result:
+
+: non existent as the Verifier is the same as the Rely Party
+
+
+## System Example Made Up of Chip Vendor and IoT Device Maker
```
mcr, line 1377: This is a second example.  Maybe we can process each example as a separate PR?

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Two entities (companies / organizations) participate in this example.
+
+The chip vendor that makes the chip. They also make the SW that is the Attester. It does the following:
+* Manufacturers the chip
+* Authors the SW that runs on the chip that is the Attester
+* Personalizes each chip with a unique 256-bit symmetric key and key ID
+* Also stores the 256-bit symmetric key in a well-secured database
+* Operates a web-based HMAC verification service the inputs to which are the key ID and HMAC to verify and the output is a yes or no on the verification
```
mcr, line 1386: I don't think that this level of detail about key generate belongs in the RATS architecture, even as an appendix.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+## System Example Base on a Third Party Verification Service
+
+Three entities (companies / organizations) participate in this example:
+
+First off, there is a device maker. It does the following:
+* Manufacturers devices using HW and SW of their choice that includes an Attester
+* Puts an attestation key into each device (symmetric or asymmetric)
+* Has the device certified by one or more certification programs
+* Generates a detailed Endorsement for each device that includes the verification key, list of certifications, reference measurement values and implicit claims
```
mcr, line 1434: This example says far less (either choice) about key generation, which validates my belief that the level of detail in the first two cases was inappropriate.

mcr (changes requested):

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+: The HW and SW made by the chip vendor
+
+Attestation Evidence:
+
+: A protocol message generated by the Attester sent to the cloud service operated by the IoT device vendor
+
+Verifier and Verifier Owner:
+
+: The Verifier and Verifier Owner are split with part run by the chip vendor and part run by the IoT device maker
```
dthaler, line 1416: This is hard to understand.  What "part" is run by each?

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Relying Party and Relying Party Owner:
+
+: are also the application vendor cloud service; since the Verify and Relying Part are the same the distinction between passport and background check doesn’t apply.
```
dthaler, line 1420:
```suggestion
: are also the application vendor cloud service; since the Verifier and Relying Party are part of the same service, this is the Background Check model.
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Attestation Result:
+
+: non existent as the Verifier is the same as the Rely Party
```
dthaler, line 1424:
```suggestion
: represented by internal APIs or state, since the Verifier is the same as the Relying Party
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+## System Example Base on a Third Party Verification Service
```
dthaler, line 1426:
```suggestion
## System Example Based on a Third-Party Verification Service
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Three entities (companies / organizations) participate in this example:
```
dthaler, line 1428:
```suggestion
Three companies or organizations participate in this example:
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+: are also the application vendor cloud service; since the Verify and Relying Part are the same the distinction between passport and background check doesn’t apply.
+
+Attestation Result:
+
+: non existent as the Verifier is the same as the Rely Party
+
+
+## System Example Made Up of Chip Vendor and IoT Device Maker
+
+Two entities (companies / organizations) participate in this example.
```
dthaler, line 1379:
```suggestion
Two companies or organizations participate in this example.
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+Attestation Result:
+
+: non existent as the Verifier is the same as the Rely Party
+
+## System Example Base on a Third Party Verification Service
+
+Three entities (companies / organizations) participate in this example:
+
+First off, there is a device maker. It does the following:
+* Manufacturers devices using HW and SW of their choice that includes an Attester
```
dthaler, line 1431:
```suggestion
* Manufactures devices using hardware and software of their choice that includes an Attester
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Puts an attestation key into each device (symmetric or asymmetric)
```
dthaler, line 1432: As I observed on the first example, this is less secure than other examples would be where the key is generated by the device itself and the private key is never seen by anyone including the manufacturer, only the public key.

draft-ietf-rats-architecture.md
dthaler, line 1434: "implicit claims" is still an undefined term

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Has the device certified by one or more certification programs
+* Generates a detailed Endorsement for each device that includes the verification key, list of certifications, reference measurement values and implicit claims
+
+The Relying Party operates some cloud service for which attestation needed, perhaps a payment service. It does the following:
+* Receives service requests from devices. The service requests includes Attestation Evidence.
```
dthaler, line 1437:
```suggestion
* Receives service requests from devices. The service requests include Attestation Evidence.
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Sends the Attestation Evidence to a 3rd party verification service.
+* Receives Attestation Results from 3rd party verification service.
```
dthaler, line 1439:
```suggestion
* Sends the Attestation Evidence to a third-party verification service.
* Receives Attestation Results from third-party verification service.
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Decides whether to grant the service, perhaps varying the payment amount allowed based on details that are in the Attestation Result such the certifications the device has received
+
+The 3rd Party Verification and Attestation Aggregation Service. It does the following:
+* Receives Attestation Evidence from the Relying Party (This uses the Background Check Model)
+* Looks up the Endorsement by an Endorsement identifier, perhaps a URI, and receives the Endorsement from the device maker, most likely over a channel that has both confidentiality, integrity protection and authentication.
+* Performs the Verification of the Attestation Evidence which involves: verifying the signature on the Attestation Evidence and comparison of claims to reference values in the Endorsement.
```
dthaler, line 1445:
```suggestion
* Performs the Verification of the Attestation Evidence which involves: verifying the signature on the Attestation Evidence and comparison of claims to reference values provided along with the Endorsement.
```

Also, if this is intended to be a complete list, it's insecure, since nothing in this sentence does a freshness check for example.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+: The device maker
+
+Endorsement:
+
+: A detailed message created by the device maker that is secured for confidentiality that contains reference values, the verification key and implicit claims
+
+Attester: 
+
+: The HW and SW made by the device maker
```
dthaler, line 1460:
```suggestion
: The hardware and software made by the device maker
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+This maps as follows in the RATS architecture:
+
+Endorser:
+
+: The device maker
+
+Endorsement:
+
+: A detailed message created by the device maker that is secured for confidentiality that contains reference values, the verification key and implicit claims
```
dthaler, line 1456: See PR #150

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Attester: 
+
+: The HW and SW made by the device maker
+
+Attestation Evidence:
```
dthaler, line 1462:
```suggestion
Evidence:
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Endorsement:
+
+: The root certificate published by the endorser
+
+Attester:
+
+: The HW made by the device OEM and the SW made by the OS vendor
+
+Attestation Evidence:
```
dthaler, line 1360:
```suggestion
Evidence:
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+: The HW and SW made by the device maker
+
+Attestation Evidence:
+
+: A protocol message generated by the Attester sent to the Relying Party
+
+Verifier and Verifier Owner:
+
+: A third-party Attestation service that can process Attestation Evidence from many different device makers and provide results to many different Relying Parties.
```
dthaler, line 1468:
```suggestion
: A third-party Attestation service that can process Evidence from many different device makers and provide results to many different Relying Parties.
```

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Relying Party and Relying Party Owner:
+
+: The operator of a cloud-based payment service.
```
dthaler, line 1472: No, the Relying Party is the payment service itself.  The operator of it is the Relying Party Owner.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+Attestation Result:
+
+: A detailed set of claims with 1) claims copied forward from the Attestation Evidence , 2) implied claims copied through from the Endorsement and 3) new claims created to report the result of comparison to reference values
```
dthaler, line 1476:
```suggestion
: A detailed set of claims with 1) claims copied forward from the Evidence, 2) implied claims copied through from the Endorsement, and 3) new claims created to report the result of comparison to reference values
```

dthaler (commented):
I agree with Michael's comments.
Also I fear that adding such examples will only result in people wanting to add 20 other examples to the doc, since even taken together they only cover a narrow subset of common waysto compose the building blocks  (e.g., there are no Passport model examples), and that's not even counting layered attestation and composite devices.  As such, just like Michael's use cases document is not in the arch doc, a separate document might be better in my view.

draft-ietf-rats-architecture.md
laurencelundblade: By "system" I mean an end-end set of entities that make up a complete function system with nothing left out. The suggestion of Smartphone doesn't work because it is only a small part of the "system". I'm OK with a different word, perhaps "End-to-end Complete Examples"?

draft-ietf-rats-architecture.md
laurencelundblade: They are  not factory generated at least not in the factory that makes the devices.  They are generated by the OS vendor, hopefully in a secure cryptographic operations center.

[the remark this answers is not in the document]
laurencelundblade: This is the only way it can work in this particular example.

I don't want to name companies, but some people are really doing this.

draft-ietf-rats-architecture.md
laurencelundblade: I think it is reasonable in this example.  Think of an OS vendor like Google supplying both the TEE which has the attester and Android.  

In some cases the TEE is made by another, but adding that would complicate the example a lot for little gain.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+This maps as follows in the RATS architecture:
+
+Endorser:
+
+: The operating system vendor is the endorser
+
+Endorsement:
+
+: The root certificate published by the endorser
```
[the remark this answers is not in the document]
laurencelundblade: I can see it either way. I picked the root because the root flows directly from the Endorser to the Verifier, where as the device's cert does not flow directly.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+# Appendix B: System Examples
+
+(Rough first draft to see if there is interest)
+
+## System Example with OS Vendor as Main Driver
+
+Three entities (companies / organizations) participate in this examples.
+
+The operating system vendor is at the center. It sets the architecture and the level of security achieved. It performs the following:
```
henkbirkholz, line 1329:
```suggestion
The operating system vendor is at the center. It sets the architecture and the level of security intended. It performs the following:
```

draft-ietf-rats-architecture.md
henkbirkholz: now I am confused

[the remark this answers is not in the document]
henkbirkholz: now I am more concerned than before

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1317,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+The device OEM does the manufacturing. It performs the following:
+* Physically makes the device to OS vendors security requirements
+* Installs the OS on the device
+* Personalizes each device with a unique private key and certificate supplied by the OS vendor
+
+The application vendor that implements some application with a back-end cloud service. It performs the following:
+* Author an app on the device that calls the attestation API to obtain Attestation Evidence
+* Convey the Attestation Evidence to the back-end cloud service over the network
+* Verify the Attestation Evidence and decided whether or not to allow application service
```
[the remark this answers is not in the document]
henkbirkholz: There seems to be a back-end cloud service. That service permits service use, if valid Evidence can be presented by the App.

henkbirkholz (commented):
very quick and small review. TBC

mcr:
We think that this belongs in a document on its own, not an unread appendix.

Issue #132, opened 2020-07-31
Trust of Attester by Verifier is transitive


laurencelundblade:
This is for section 7.2 on Attester, 7.4 on Verifier and 7.5 the Endorser/Manufacturer.

The main trust flow for all of attestation is that the Verifier has a direct trust relationship with the Endorser/Manufacturer but not the Attester. The Endorser/Manufacturer supplies Input (e.g., an Endorsement containing key material) to the Verifier that allows the Verifier to come to trust the Attester. 

In just about every use of attestation, there is no direct trust of the Attester by the Verifier. The Verifier doesn't directly trust the Attester's HW or SW as suggested by the text in 7.4. This text seems incorrect on one of the most fundamental attestation concepts.

Section 7.2 and 7.5 don't even discuss the primary transitive trust flow. They only address the secondary trust flow on confidentiality and privacy. While confidentiality and privacy are important they are not the primary reason that the Attester and Endorser exist.

mcr:
closed by #154.

PR #149, opened 2020-09-17
Revise description of attester<->verifier trust

laurencelundblade:
This adds a paragraph on the primary trust flow to the Attester sub section.  Previously that subsection only discussed secondary trust flow (confidentiality flow)

I'm pretty sure that just about every real attestation use involves the Verifier coming to trust the Attester by receipt of a nonce (or equivalent) signed or authenticated by a key in and protected by the Attester. I don't see how a Verifier can trust an Attester in any other way for just about every attestation use case. I've revised the text to say that.

I think all the discussion about ROM code and HW defenses is now covered in Security Consideration.

The addresses #132


draft-ietf-rats-architecture.md
```diff
@@ -718,6 +718,13 @@ could establish trust in a Verifier as discussed above.
 
 ## Attester
 
+The Attester proves it is trustworthy to the Verifier by using key material provisioned into it by the Endorser.
```
dthaler, line 721: "provisioned into it by the Endorser" is not true in general.   For example, it's not true in the case where the Attester is an IoT device that generates its own key pair at the factory, and the public key is then extracted and used to generate a certificate issued by an Endorser (the manufacturer or whoever else), and the certificate itself isn't provisioned on the device but can be looked up on demand by the Verifier.  In this case, there is no "key material" provisioned into the Attester by any entity other than the Attester itself.

Line 723 and 754 are incorrect for the same reason.

draft-ietf-rats-architecture.md
```diff
@@ -718,6 +718,13 @@ could establish trust in a Verifier as discussed above.
...
+The proof is by the signing or authenticating of Attestation Evidence (see discussion on freshness elsewhere in this document).
```
dthaler, line 722:
```suggestion
The proof is by the signing or authenticating of Evidence (see {{freshness}}).
```

draft-ietf-rats-architecture.md
```diff
@@ -718,6 +718,13 @@ could establish trust in a Verifier as discussed above.
...
+The Endorser only puts this key material into Attesters that it trusts.
+This part of the Attester is often known as, or performed by, a root of trust.
+The security strength or value of the overall Attestation depends greatly on the security strength of the implementation of the root of trust.
+See further discussion in Security Considerations.
```
dthaler, line 726: This would need to be a markdown cross reference.

draft-ietf-rats-architecture.md
```diff
@@ -743,19 +750,9 @@ in order to bootstrap the sequence.
...
-might appraise the trustworthiness of an application component, operating
-system component, or service under the assumption that information
-provided about it by the lower-layer hypervisor or firmware is true.
-A stronger level of assurance of security comes when information can be vouched
-for by hardware or by ROM code, especially if such hardware is
-physically resistant to hardware tampering.  The component that is
-implicitly trusted is often referred to as a Root of Trust.
+The Verifier trusts (or more specifically, the Verifier's security policy is written in a way that configures the Verifier to trust) an Endorser who supplies an Endorsement which contains information by which the Verifier can come to trust the Attester.
+Typically, that information is a key used to verify the signature on Attestation Evidence which proves that the Attester has the private key that is only put into trustworthy devices by the Endorser.
+That is, the Verifier comes to trust the Attester transitively through the Endorser.
```
dthaler, line 755: Hard to review since line breaks were removed.  Please revert line breaks.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
 valid.  Thus, if the Attestation Result sends a validity lifetime
 in terms of time(RX)-time(RG), then the Relying Party can check
 `time(OP) - time(ER) < time(RX)-time(RG)`.
+
+# Appendix B: System Examples
+
+(Rough first draft to see if there is interest)
+
```
dthaler, line 1321: Keep examples in a separate PR

dthaler (changes requested):

draft-ietf-rats-architecture.md
laurencelundblade: Yes. Github mistake.

draft-ietf-rats-architecture.md
laurencelundblade: We probably should come up with better phrase -- "establish key material" or "set up key material" or "personalization"? It seems awkward to explain all the options each time it needs to be mentioned.

Even if the key material is generated on the device, you still have to do real work to 1) provide integrity when the public key is coming off the device and 2) make sure you are only taking keys off of real devices (no imposter, forgery or midnight runs). 

I have seen the term "provision" used to mean the whole process at a high level.

I know you keep bringing this up. I very much agree that what you suggest is the best way and should be allowed for and encouraged in the architecture doc, but I also know that are many situations where it can't be done. For example, it can't be done with symmetric keys. It can be done when you are putting keys into fuses on a chip that has no HW key generation feature. It can't be done when keys are supplied by someone other than the manufacturer.

So it seems we need some term or wording to describe this that covers all the scenarios.

henkbirkholz: A common problem.

Maybe: "Attesters prove their trustworthiness to Verifiers via their key material that is established in their Attesting Environments under the control and authority of a trusted Endorsers." (might have an effect on the following sentences)

draft-ietf-rats-architecture.md
```diff
@@ -718,6 +718,16 @@ could establish trust in a Verifier as discussed above.
 
 ## Attester
 
+Attesters prove their trustworthiness to Verifiers via key material that is established in their Attesting Environments by an Attester manufacturer and/or an Endorser.
+The proof is by the signing or authenticating of Evidence. See {{freshness}}.
+The Attester manufacturer and/or Endorser only puts this key material into Attesters that it trusts.
+This part of the Attester is often known as, or performed by, a root of trust.
+The security strength or value of the overall Attestation depends greatly on the security strength of the implementation of the root of trust.
+See {{security-considerations}} for discussion on security strength.
```
mcr, line 726:
```suggestion

```

draft-ietf-rats-architecture.md
```diff
@@ -718,6 +718,15 @@ could establish trust in a Verifier as discussed above.
 
 ## Attester
 
+Attesters prove their trustworthiness to Verifiers via key material that is established in their Attesting Environments by an Attester manufacturer and/or an Endorser.
+The proof is by the signing or authenticating of Evidence. See {{freshness}}.
```
dthaler, line 722:
```suggestion
Attesters prove their trustworthiness to Verifiers via Evidence signed using key material that is established in their Attesting Environments, and some way of ensuring freshness (see {{freshness}}).
```

draft-ietf-rats-architecture.md
nedmsmith: Other than line breaks, I agree with the new content.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+(Rough first draft to see if there is interest)
+
+## System Example with OS Vendor as Main Driver
+
+Three entities (companies / organizations) participate in this examples.
+
+The operating system vendor is at the center. It sets the architecture and the level of security achieved. It performs the following:
+* Generates large quantities of ECDSA key pairs to be used as per-device attestation keys
+* Signs the public keys using an attestation PKI and root that they control creating per-device certificates.
```
nedmsmith: Need to disambiguate 'root' as it could refer to a root CA or a root of trust component.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+# Appendix B: System Examples
+
+(Rough first draft to see if there is interest)
+
+## System Example with OS Vendor as Main Driver
+
+Three entities (companies / organizations) participate in this examples.
+
+The operating system vendor is at the center. It sets the architecture and the level of security achieved. It performs the following:
```
nedmsmith: It isn't clear which attestation role the operating system vendor is assuming. As an endorser that provisions the attestation private key and the mfg CA that issues the attestation ID certificate, this often isn't an operating system vendor. Since this is an example, the vendor very well could be, but that isn't usually how it works.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
 valid.  Thus, if the Attestation Result sends a validity lifetime
 in terms of time(RX)-time(RG), then the Relying Party can check
 `time(OP) - time(ER) < time(RX)-time(RG)`.
+
+# Appendix B: System Examples
```
nedmsmith: Was there consensus that this belongs in the architecture draft?

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+(Rough first draft to see if there is interest)
+
+## System Example with OS Vendor as Main Driver
+
+Three entities (companies / organizations) participate in this examples.
+
+The operating system vendor is at the center. It sets the architecture and the level of security achieved. It performs the following:
+* Generates large quantities of ECDSA key pairs to be used as per-device attestation keys
+* Signs the public keys using an attestation PKI and root that they control creating per-device certificates.
+* Securely sends the private keys and certificates to the device OEM in batches to be used in manufacturing
```
nedmsmith: It is difficult to follow who the subject of the sentence is. If it is the operating system vendor from line 1326 then it doesn't make sense for the OS vendor to know private keys for attesting environments. This draft shouldn't give the impression that it advocates doing that.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Authors the device SW that is the Attester and ships it to the OEMs. This is a part of the OS.
+* Publishes the root certificate and definition of the claims
+
+The device OEM does the manufacturing. It performs the following:
+* Physically makes the device to OS vendors security requirements
+* Installs the OS on the device
+* Personalizes each device with a unique private key and certificate supplied by the OS vendor
```
nedmsmith: I expected the OEM would be in charge of the key creation and certificate issuance since it is a use case where the OEM is 'driving'.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+This maps as follows in the RATS architecture:
+
+Endorser:
+
+: The operating system vendor is the endorser
+
+Endorsement:
+
+: The root certificate published by the endorser
```
nedmsmith: Unclear if root refers to a root CA or root of trust / Attesting Environment.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+: non existent as the Verifier is the same as the Rely Party
+
+
+## System Example Made Up of Chip Vendor and IoT Device Maker
+
+Two entities (companies / organizations) participate in this example.
+
+The chip vendor that makes the chip. They also make the SW that is the Attester. It does the following:
+* Manufacturers the chip
```
nedmsmith: manufactures - typo?

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+* Authors the SW that runs on the chip that is the Attester
+* Personalizes each chip with a unique 256-bit symmetric key and key ID
+* Also stores the 256-bit symmetric key in a well-secured database
+* Operates a web-based HMAC verification service the inputs to which are the key ID and HMAC to verify and the output is a yes or no on the verification
```
nedmsmith: Is the attestation evidence in the form of an HMAC? Is the Verifier supplying a query to the mfr'r web service in the form of an HMAC and key ID?

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+The IoT device maker buys chips from the vendor, makes the chip the core of an IoT device (perhaps sensors for a home security service). It does the following:
+* Manufacturers and a device that an end user may purchase
+* Adds their own SW to the device over the top of the SW the chip vendor supplied
```
nedmsmith: This deployment case seems to be 80% overlapping the previous one.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+This maps as follows in the RATS architecture:
+
+Endorser:
+
+: The device maker
+
+Endorsement:
+
+: A detailed message created by the device maker that is secured for confidentiality that contains reference values, the verification key and implicit claims
```
nedmsmith: we don't define the term implicit claims (yet) therefore needs some explanation here.
Also need to resolve if "Endorser" can supply "reference values" or if only "Reverence Value Provider" does that.

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+: A third-party Attestation service that can process Attestation Evidence from many different device makers and provide results to many different Relying Parties.
+
+Relying Party and Relying Party Owner:
+
+: The operator of a cloud-based payment service.
+
+Attestation Result:
+
+: A detailed set of claims with 1) claims copied forward from the Attestation Evidence , 2) implied claims copied through from the Endorsement and 3) new claims created to report the result of comparison to reference values
```
nedmsmith: I assume 'implied claims' are the same as 'implicit claims' and 'static claims'?

draft-ietf-rats-architecture.md
```diff
@@ -1317,3 +1314,160 @@ use beyond the period for which it deems the Attestation Result to remain
...
+
+: A detailed message created by the device maker that is secured for confidentiality that contains reference values, the verification key and implicit claims
+
+Attester: 
+
+: The HW and SW made by the device maker
+
+Attestation Evidence:
+
+: A protocol message generated by the Attester sent to the Relying Party
```
nedmsmith: presumably the protocol message contains Evidence?

draft-ietf-rats-architecture.md
```diff
@@ -718,6 +718,15 @@ could establish trust in a Verifier as discussed above.
 
 ## Attester
 
+Attesters prove their trustworthiness to Verifiers via key material that is established in their Attesting Environments by an Attester manufacturer and/or an Endorser.
+The proof is by the signing or authenticating of Evidence. See {{freshness}}.
+The Attester manufacturer and/or Endorser only puts this key material into Attesters that it trusts.
+This part of the Attester is often known as, or performed by, a root of trust.
+The security strength or value of the overall Attestation depends greatly on the security strength of the implementation of the root of trust.
+
+There is also a non-remote attestation use case where the Attester and Verifier are directly connected such that cryptographic proof of trustworthiness is unnecessary (e.g., they are on the same chip).
```
henkbirkholz, line 727: I am not comfortable wit using the phrase "non-remote attestation". The reader has no reference or context here and I find it quite confusing.

henkbirkholz:
In general, I find this PR very hard to discuss with its current scope. There seem to be very specific issues with the text that are discussed in this PR right now. And then there seems to be no agreement about if this proposal is deemed useful in general.

Could we please streamline this? (break at any given time, if the answer to a question is no)

1.) Get an agreed upon understanding, if this PR is worthwhile to be addressed
2.) Identify definable issues with the current text this PR modifies and re-issue smaller-scoped PRs

Alternatively, please bash this proposal via your own :-)

draft-ietf-rats-architecture.md
henkbirkholz: We should avoid "implied" or "implicit" entirely and find out if "vouched for" suits the text better here.

draft-ietf-rats-architecture.md
henkbirkholz: pls see the comment below

PR #150, opened 2020-09-18
Update Reference Values discussion

dthaler:
Need Peter to confirm he agrees, but I believe these changes are consistent with the points that Peter and I were both making during the last call.


mcr (approved):

William-PanWei (approved):

draft-ietf-rats-architecture.md
```diff
@@ -166,6 +166,15 @@ Evidence:
...
 
+Reference Value Provider:
+
+: An entity (typically a manufacturer) whose Reference Values help Verifiers appraise the authenticity of Evidence.
+
+Reference Values:
+
+: A set of values against which values of Claims can be compared as part of
+applying an Appraisal Policy for Evidence.
+
```
laurencelundblade, line 177: Add "also called known-good values"

draft-ietf-rats-architecture.md
```diff
@@ -823,7 +834,7 @@ private key is only known inside the device's hardware.  Thus, when Evidence and
 are used together, an appraisal procedure can be conducted based on Appraisal Policies that may not be specific to the
 device instance, but merely specific to the manufacturer providing the Endorsement. For example,
 an Appraisal Policy might simply check that devices from a given manufacturer have information
-matching a set of known-good reference values, or an Appraisal Policy might have a set of more complex
+matching a set of known-good Reference Values, or an Appraisal Policy might have a set of more complex
```
laurencelundblade, line 837: Remove "known-good", assuming "known-good" is mentioned as an AKA for Reference Values.

draft-ietf-rats-architecture.md
```diff
@@ -851,7 +862,7 @@ An Attestation Result that indicates compliance can be used by a Relying Party t
 authorization decisions based on the Relying Party's Appraisal Policy.  The simplest such
 policy might be to simply authorize any party supplying a compliant Attestation Result
 signed by a trusted Verifier.  A more complex policy might also entail comparing information
-provided in the result against known-good reference values, or applying more complex logic
+provided in the result against known-good Reference Values, or applying more complex logic
```
laurencelundblade, line 865: Remove "known-good", assuming "known-good" is mentioned as an AKA for Reference Values.

laurencelundblade (changes requested):
Was a little surprised by this since previous consensus was that Ref Vals were either part of an Endorsement or Policy, but am entirely happy with this. It really gives clarity.

laurencelundblade:
To complete this, a sub section for Reference Value Provider should be added to section 7 on the Trust Model.  All other such entities are described there now.

henkbirkholz (changes requested):
As Laurence already indicated via comment here:

Adding a small list of common synonyms for reference values would really help here, I think.

known good values (KGV), golden measurements, nominal values

draft-ietf-rats-architecture.md
dthaler: Done

draft-ietf-rats-architecture.md
dthaler: Expanded definition

draft-ietf-rats-architecture.md
dthaler: Done

Issue #136, opened 2020-08-03
Endorser trust, section 7.5


laurencelundblade:
7.5 says that the way for an Endorser to trust and Verifier is to use attestation.

While I won't categorically say this should never be done, it seems like this is overly complex and a somewhat misuse of attestation.

I think of the Verifier as a service like a web site, not an implementation like a TEE, so TLS or similar, not attestations, seems like the more efficient and conventional way to do this.  

I think attestation is associated with actual HW and SW, for example the Dell server, Intel CPU and Web stack used to run the service. The operator of a Verifier gets to change those around without having to tell the people using the service.

On the other hand, if a Verification Service changes owners, say from Chinese to American, then Endorser concerned about confidential stuff in the Endorsements really wants to know.

I also think that Verifier Owner should not be lumped in with Endorser. The Verifier Owner is very likely to be tightly coupled organizationally with the Verifier, where the Endorser is not likely to be.

laurencelundblade:
PR #153 addresses this well enough.

PR #153, opened 2020-09-24
TLS authentication for secondary trust

laurencelundblade:
TLS authentication is sufficient for establishing trust in the various entities before sending them privacy sensitive data. TLS is also widely supported today and attestation between servers is not. Most likely TLS will be the mechanism. This PR adds mention to in addition to attestation.

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -734,6 +734,7 @@ For example, one explicit way for a Relying Party "A" to establish
 such trust in a Verifier "B", would be for B to first act as an Attester
 where A acts as a combined Verifier/Relying Party.  If A then accepts B as
 trustworthy, it can choose to accept B as a Verifier for other Attesters.
+TLS authentication of the Verifier by the Relying Party is also a possible solution.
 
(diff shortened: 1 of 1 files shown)


laurencelundblade:
Addresses #136

mcr (approved):
Looks good for me.

draft-ietf-rats-architecture.md
```diff
@@ -734,6 +734,7 @@ For example, one explicit way for a Relying Party "A" to establish
 such trust in a Verifier "B", would be for B to first act as an Attester
 where A acts as a combined Verifier/Relying Party.  If A then accepts B as
 trustworthy, it can choose to accept B as a Verifier for other Attesters.
+TLS authentication of the Verifier by the Relying Party is also a possible solution.
```
mcr, line 737:
```suggestion

The Relying Party can establish trust in the Verifier by out of band establishment of key-material, combined with a protocol like TLS to communicate. 
There is an assumption that between the establishment of the trusted key-material, and the creation of the Evidence, that the Verifier has not been compromised.
```

Issue #163, opened 2020-10-30
Architecture document readability flow


SimonFrost-Arm:
As part of a review of the document by the Arm team, the document was shown to someone less familiar with the area. Their feedback was that they found the flow of the document hard to follow, with the comments that the terminology section was too remote from the core architecture discussion. Further, within the Terminology section itself the alphabetic layout was not helpful as they had to go back and forth as they tried to understand the terms. The suggestion was to use a more hierarchical layout for the terminology.

SimonFrost-Arm:
Note that a proposed PR to address this issue has been created as https://github.com/ietf-rats-wg/architecture/pull/164

PR #164, opened 2020-10-30
readability rework for use case and terminology sections

SimonFrost-Arm:
Proposed rework of the document flow to address the readability issue raised here: https://github.com/ietf-rats-wg/architecture/issues/163.
Summary of changes: 
Move the 'scene setting' Use Cases section to follow the Introduction
Remove specific terminology (Attester / RP) from the Use Cases section by generalising the language used, while keeping the clarification of those parties
Terminology section now follows Use Case section, which puts it next to the Architecture section, and in particular Fig1, where understanding of the terminology is most needed.
Change alphabetic layout of Terminology section to be more hierarchical on use/importance of terms
Add relationship descriptions between terms.

I know that readability is subjective, but this flow has tested well in local review!


henkbirkholz (commented):
This PR mixes proposals for two issues, I think.

1.) Moving (and reordering) the Terminology session is of course to be discussed. In comparison, it seems to me the smaller issue addressed. Please mind, that moving the terminology section to the top was an intentional action with the intent to improve readability. The ordering is more a question of style - and if semantic/hierarchic ordering helps, I am all for it.

If your activity of using this document during implementation proves otherwise, that most definitely has to be taken under consideration.

2.) Changing the layout of use cases as you proposed resonates well with my personal perception. Alas, adding well-defined terms here to help with the mapping to the terminology - therefore straying from the defining principle that use cases deliberately do not use any well-defined words, but colloquial language - was also a conscious decision. I am quite okay with your proposal, as it aligns with the principle of least surprised. But, I can also see the point that early mapping can help. Hence, this also has to be discussed.

Could you please summarize both pieces of your proposal on rats@ietf.org and annotate them with your reasoning to start that discussion?

mcr:
> This PR mixes proposals for two issues, I think.

It seems like two pull requests would be easier for us to come to consensus on.

draft-ietf-rats-architecture.md
```diff
@@ -249,13 +179,7 @@ The system components perform a series of measurements that may be
...
 firmware, BIOS, software, etc. that is running.
 
-Attester:
-
-: A device desiring access to a network
-
-Relying Party:
-
-: A network infrastructure device such as a router, switch, or access point
+In this case, the device desiring access to a network must produce attestation for consideration by a network infrastructure device such as a router, switch, or access point
```
dthaler, line 182: Not sure I like "produce attestation" as a phrase.  First, it would require "attestation" to be defined in a way that is different from any other use in the document.  Second, it also implies that the device (Attester) actually produce something that is consumed by a router (Relying Party), which is not in the case in this architecture... the Attester produces Evidence for consideration by a Verifier, which produces Attestation Results for consideration by Relying Parties.

draft-ietf-rats-architecture.md
```diff
@@ -249,13 +179,7 @@ The system components perform a series of measurements that may be
 signed by the root of trust, considered as Evidence about the hardware,
```
dthaler, line 179: This PR is problematic in that it starts using capitalized terms like Evidence before being defined

draft-ietf-rats-architecture.md
```diff
@@ -270,13 +194,7 @@ that can assess its trustworthiness.  If remote attestation succeeds,
...
 a model the Attester already has in encrypted form, to the requester.
 
-Attester:
-
-: A device desiring to run an ML model
-
-Relying Party:
-
-: A server or service holding ML models it desires to protect
+In this case, device desiring to run an ML model must produce attestation for consideration by a server or service holding ML models it desires to protect
```
dthaler, line 197:
```suggestion
In this case, the device desiring to run an ML model must produce attestation for consideration by a server or service holding ML models it desires to protect
```

draft-ietf-rats-architecture.md
```diff
@@ -372,13 +261,105 @@ The FIDO protocol supports several remote attestation protocols and a mechanism
...
+Used by: Verifier;
+Produced by: Verifier Owner.
+
+
+Appraisal Policy for Attestation Results:
+
+: A set of rules that direct how a Relying Party uses the Attestation Results regarding an Attester generated by the Verifiers. Compare /security policy/ in {{RFC4949}}.
+Used by: Relying Party;
+Produced by: Relying Party Owner.
+
```
dthaler, line 349: Another problem is that this PR deletes the definition of Endorsements from line 178 and doesn't add it back into this terminology section.

draft-ietf-rats-architecture.md
```diff
@@ -372,13 +261,105 @@ The FIDO protocol supports several remote attestation protocols and a mechanism
...
+# Terminology {#terminology}
+
+## Roles
+
 Attester:
 
-: Every FIDO Authenticator contains an Attester.
+: A role performed by an entity (typically a device) whose Evidence must be appraised in order to infer the extent to which the Attester is considered trustworthy, such as when deciding whether it is authorized to perform some operation.
+Produces: Evidence.
```
dthaler, line 276: I do kind of like the Produces: and Consumes: annotation.  I'd be fine with adding such into the terminology section if others agree.

This PR does make it hard to see changes to the terminology section though, I'd prefer to see a separate PR for such things.

dthaler (changes requested):

draft-ietf-rats-architecture.md
SimonFrost-Arm: Evidence is briefly defined in the introduction as well as more formally covered in the Terminology section (also Attester and Relying Party).

PR #228, opened 2020-12-08
add small clarification to Example 1

### LINKED ISSUE #225: 16.1. Example 1, troubles with the third paragraph
Section "16.1. Example 1", third para:
```
If time(VG_a) is also included in the Evidence along
with the claim value generated at that time, and the
Verifier decides that it can trust the time(VG_a) value,
the Verifier can also determine whether the claim value
is recent by checking time(RG_v) - time(VG_a) < Threshold,
again where the threshold is large enough to account for
the maximum permitted clock skew between the Verifier
and the Attester.
```
I find the reference to clock skew slightly confusing.

The Threshold value here is more likely to depend on a verification policy
that defines what is an acceptable "age" of the sampled claim rather than
on considerations about clocks drifting.

thomas-fossati:
Clarify that recentness of Evidence depends primarily on the appraisal policy and that clock skew is only secondary.

Fixes #225

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1312,8 +1312,9 @@ permitted clock skew between the Verifier and the Attester.
 If time(VG_a) is also included in the Evidence along with the claim value
 generated at that time, and the Verifier decides that it can trust the
 time(VG_a) value, the Verifier can also determine whether the claim value is
-recent by checking `time(RG_v) - time(VG_a) < Threshold`, again where the threshold
-is large enough to account for the maximum permitted clock skew between
+recent by checking `time(RG_v) - time(VG_a) < Threshold`, where the threshold
+is decided by the Appraisal Policy for Evidence, and again needs to take
+into account the maximum permitted clock skew between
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -1312,8 +1312,9 @@ permitted clock skew between the Verifier and the Attester.
 If time(VG_a) is also included in the Evidence along with the claim value
 generated at that time, and the Verifier decides that it can trust the
 time(VG_a) value, the Verifier can also determine whether the claim value is
-recent by checking `time(RG_v) - time(VG_a) < Threshold`, again where the threshold
-is large enough to account for the maximum permitted clock skew between
+recent by checking `time(RG_v) - time(VG_a) < Threshold`, where the threshold
+is decided by the Appraisal Policy for Evidence, and again needs to take
```
mcr, line 1316:
```suggestion
recent by checking `time(RG_v) - time(VG_a) < Threshold`.  
The threshold is decided by the Appraisal Policy for Evidence, and again needs to take
```

PR #240, opened 2020-12-15
drop OPC UA in favour of DICE and TLS-CWT

### LINKED ISSUE #236: the "Claims Encoding Formats" back-story needs some work
This sentence:
```
   To enable remote attestation to be added to existing protocols,
   enabling a higher level of assurance against malware for example, it
   is important that information needed for appraising the Attester be
   usable with existing protocols that have constraints around what
   formats they can transport.
```
it's quite hard to digest.

But more generally, the motivating use case (OPC UA) is slightly convoluted, and I'm wondering whether it is really needed?  But assuming it were, the exposition needs to be improved to present a clear and compelling case.

thomas-fossati:
Fixes #236

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -85,6 +85,15 @@ normative:
 informative:
   RFC4949:
   RFC8322:
+  OPCUA:
+    author:
+      org: "OPC Foundation"
+    title: "OPC Unified Architecture Specification, Part 2: Security Model, Release 1.03"
+    date: 2015-11-25
+    target: https://opcfoundation.org/developer-tools/specifications-unified-architecture/part-2-security-model/
+    seriesinfo:
+      OPC 10000-2
+
   TCG-DICE:
     author:
       org: "Trusted Computing Group"
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1007,7 +1007,10 @@ The following diagram illustrates a relationship to which remote attestation is 
 In this diagram, the protocol between Attester and a Relying Party
 can be any new or existing protocol (e.g., HTTP(S), COAP(S),
 ROLIE {{RFC8322}},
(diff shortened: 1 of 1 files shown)


mcr (approved):

draft-ietf-rats-architecture.md
```diff
@@ -1001,24 +998,18 @@ The following diagram illustrates a relationship to which remote attestation is
...
-For example, OPC UA {{OPCUA}} (probably the most common protocol in
-industrial IoT environments) is defined to carry X.509 certificates and so
-security information must be embedded into an X.509 certificate to be passed
-in the protocol.  Thus, remote attestation related information could be natively
-encoded in X.509 certificate extensions, or could be natively encoded in
-some other format (e.g., a CWT) which in turn is then encoded in an X.509
-certificate extension.
-
-Especially for constrained nodes, however, there is a desire to minimize
+802.1x, OPC UA, etc.), depending on the use case.
```
dthaler, line 1001: If we drop OPC UA below, we should drop it from this line too.

mcr, line 1001:
```suggestion
802.1x, OPC UA {{OPCUA}}, etc.), depending on the use case.
```

draft-ietf-rats-architecture.md
```diff
@@ -85,14 +85,11 @@ normative:
 informative:
   RFC4949:
   RFC8322:
-  OPCUA:
+  TCG-DICE:
```
mcr, line 88:
```suggestion
OPCUA:
    author:
      org: "OPC Foundation"
    title: "OPC Unified Architecture Specification, Part 2: Security Model, Release 1.03"
    date: 2015-11-25
    target: https://opcfoundation.org/developer-tools/specifications-unified-architecture/part-2-security-model/
    seriesinfo:
      OPC 10000-2

  TCG-DICE:
```

PR #241, opened 2020-12-15
small fix in the "Combinations" section

### LINKED ISSUE #234: possible imprecision in the "Combinations" section
We say:
```
   It is also worth pointing out that the choice of model is generally
   up to the Relying Party
```

I'm not sure it "is up to the RP"_.  Maybe better saying it "depends on the RP application" (or similar) because it's typically not a choice of the RP but of the use case / protocol flow in which the RP operates...

thomas-fossati:
Fixes #234

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -723,7 +723,7 @@ One variation of the background-check model is where the Relying Party
 and the Verifier are on the same machine, performing both functions together. 
 In this case, there is no need for a protocol between the two.
 
-It is also worth pointing out that the choice of model depends on the use case, and that different applications may use different models while still interacting with the same Attester or Verifier.
+It is also worth pointing out that the choice of model depends on the use case, and that different Relying Parties may use different topological patterns.
 
 The same device may need to create Evidence for different Relying Parties and/or different use cases. 
 For instance, it would provide Evidence to a network infrastructure device to gain access to the network, and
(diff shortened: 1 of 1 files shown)


mcr (approved):

draft-ietf-rats-architecture.md
```diff
@@ -723,7 +723,8 @@ One variation of the background-check model is where the Relying Party
 and the Verifier are on the same machine, performing both functions together. 
 In this case, there is no need for a protocol between the two.
 
-It is also worth pointing out that the choice of model is generally up to the Relying Party. 
+It is also worth pointing out that the choice of model depends on the use case, and that different applications may use different models while still interacting with the same Attester or Verifier.
```
dthaler, line 726: I disagree with this wording, which implies that the term "application" inherently implies a Relying Party and not an Attester.
```suggestion
It is also worth pointing out that the choice of model depends on the use case, and that different Relying Parties may use different models while still interacting with the same Attester or Verifier.
```

dthaler: Actually I'm not sure the updated statement is actually true.   If you have a Relying Party that expects Evidence because the Relying Party is using the background check model, then it won't work with an Attester that only sends it Attestation Results and not the Evidence (or claims from the evidence) since the RP's Verifier won't receive the claims it needs to appraise.

I think the statement is more correct in the original and this change should be rejected.

PR #248, opened 2021-01-11
Improve clarity of the handle example

### LINKED ISSUE #198: Handles appendix needs some care
The appendix on Handles needs some work to better clarify the concepts and simplify the flow.

Some things about the text:

* Clearly define what is a handle;
* Explain its relation with epoch establishment;
* Clarify network jitter impact on handle validity;
* Explain what are the base rules for working with handles;

Some further points on the sequence diagram:

* Fix confusing legend;
* Add some accompanying prose to explain the depicted use case as it's not self-evident;

thomas-fossati:
Fixes #198

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1491,8 +1491,8 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
 
 ## Example 3: Handle-based Passport Model Example
 
-The example in {{fig-handle-passport}} illustrates a Handle-based interaction
-using the passport topology model.
+The example in {{fig-handle-passport}} illustrates a hypothetical Passport Model
+solution that uses handles instead of nonces or timestamps.
 
 The Handle Distributor broadcasts handle `H` which starts a new
 epoch `E` for a protocol participant upon reception at `time(HR)`.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1076,7 +1076,7 @@ a different format.
 ~~~~
 {:multievidence #multievidence_diag title="Multiple Attesters and Relying Parties with Different Formats"}
 
-# Freshness
+# Freshness {#freshness}
 
 A Verifier or Relying Party may need to learn the point in time
 (i.e., the "epoch") an Evidence or Attestation Result has been produced.  This
@@ -1341,7 +1341,8 @@ or might be defined relative to some other timestamp or timeticks counter.
 | VG | Value generated             | A value to appear in a Claim was created.  In some cases, a value may have technically existed before an Attester became aware of it but the Attester might have no idea how long it has had that value.  In such a case, the Value created time is the time at which the Claim containing the copy of the value was created.
 | NS | Nonce sent                  | A nonce not predictable to an Attester (recentness & uniqueness) is sent to an Attester.
 | NR | Nonce relayed               | A nonce is relayed to an Attester by another entity.
-| HR | Handle received             | A handle distributed by a Handle Distributor was received.
+| HT | Handle transmitted          | A new handle is transmitted by a Handle Distributor to instruct the receiving entities to transition to a new epoch.
+| HR | Handle received             | A handle is successfully received and processed by an entity.
 | EG | Evidence generation         | An Attester creates Evidence from collected Claims.
 | ER | Evidence relayed            | A Relying Party relays Evidence to a Verifier.
 | RG | Result generation           | A Verifier appraises Evidence and generates an Attestation Result.
@@ -1492,20 +1493,29 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
 
 ## Example 3: Handle-based Passport Model Example
 
-The following example illustrates a hypothetical Passport Model
-solution that uses handles instead of nonces or timestamps.
-In this example, Evidence generation based on received handles always uses the current (most recent) handle.
-As handles are distributed over the network, all involved entities receive a fresh handle at roughly the same time.
-Due to distribution over the network, there is some jitter with respect to the time the Handle is received, time(HR), for each involved entity.
-To compensate for this jitter, there is a small period of overlap (a specified offset) in which both a current handle and corresponding former handle are valid in Evidence appraisal: `validity-duration = time(HR'_v) + offset - time(HR_v)`. The offset is typically based on a network's round trip time.
-Analogously, the generation of valid Evidence is only possible, if the age of the handle used is lower than the validity-duration: `time(HR_v) - time(EG_a) < validity-duration`.
+The example in {{fig-handle-passport}} illustrates a Handle-based interaction
+using the passport topology model.
 
-From the point of view of a Verifier, the generation of valid Evidence is only possible, if the age of the handle used in the Evidence generation is younger than the duration of the distribution interval -- "(time(HR'_v)-time(HR_v)) - (time(HR_a)-time(EG_a)) \< validity-duration".
+The Handle Distributor broadcasts handle `H` which starts a new
+epoch `E` for a protocol participant upon reception at `time(HR)`.
 
-Due to the validity-duration of handles, multiple different pieces of Evidence can be generated based on the same handle.
-The resulting granularity (time resolution) of Evidence freshness is typically lower than the resolution of clock-based tickcounters.
+The Attester generates Evidence incorporating handle `H` and conveys it to the
+Verifier.
 
-The following example illustrates a hypothetical Background-Check Model solution that uses handles and requires a trustworthy time source available to the Handle Distributor role.
+The Verifier appraises that the received handle `H` is "fresh" according to the
+definition provided in {{freshness}}, and generates an Attestation Result.  The
+Attestation Result is conveyed to the Attester.
+
+After the transmission of handle `H'` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
+
+In the illustrated scenario, the Handle for relaying an Attestation Result to
+the Relying Party is fresh, while the handle that was used to generate the
+handle is stale. This indicates that at least one epoch has passed in between,
+which can be taken into account in Appraisal Policies for Attestation Results
+about freshness.
 
 ~~~~
                   .-------------.
@@ -1516,27 +1526,23 @@ The following example illustrates a hypothetical Background-Check Model solution
         |                |                |               |
         ~                ~                ~               ~
         |                |                |               |
-     time(HR_a)<---------+-------------time(HR_v)------>time(HR_r)
+     time(HR_a)<---------+-----------time(HR_v)----->time(HR_r)
         |                |                |               |
      time(EG_a)          |                |               |
-        |----Evidence{time(EG_a)}-------->|               |
-        | {Handle1,time(EG_a)-time(VG_a)}|                |
+        |--E{H,time(VG_a)}--------------->|               |
+        |                |                |               |
         |                |             time(RG_v)         |
-        |<-----Attestation Result---------|               |
-        |   {time(RG_v),time(RX_v)}       |               |
-        |                |                                |
-        ~                ~                                ~
-        |                |                                |
-     time(HR_a')<--------'---------------------------->time(HR_r')
-        |                                                 |
-     time(RR_a)                                           /
-        |--Attestation Result{time(RX_v)-time(RG_v)}-->time(RA_r)
-        |    {Handle2, time(RR_a)-time(EG_a)}             |
-        ~                                                 ~
-        |                                                 |
-        |                                              time(OP_r)
-        |                                                 |
+        |<----------------R{H,time(RX_v)}-|               |
+        |                |                |               |
+     time(HR'_a)<--------+---------time(HR'_v)---->time(HR'_r)
+        |                |                |               |
+     time(RR_a)          |                |               |
+        |--RR{H',R{H,time(RX_v)}}--------------------->time(RA_r)
+        |                |                |               |
+        ~                ~                ~               ~
+        |                |                |               |
 ~~~~
+{: #fig-handle-passport title="Handle-based Passport Model"}
 
 ## Example 4: Timestamp-based Background-Check Model Example
 
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1391,7 +1391,6 @@ timestamp can directly compare it to its own clock and timestamps.
         ~                                                 ~
         |                                                 |
         |                                              time(OP_r)
-        |                                                 |
 ~~~~
 
 In the figures above and in subsequent sections, curly braces
@@ -1529,18 +1528,21 @@ about freshness.
      time(HR_a)<---------+-----------time(HR_v)----->time(HR_r)
         |                |                |               |
      time(EG_a)          |                |               |
-        |--E{H,time(VG_a)}--------------->|               |
+        |---Evidence{H,time(VG_a)}------->|               |
         |                |                |               |
-        |                |             time(RG_v)         |
-        |<----------------R{H,time(RX_v)}-|               |
+        |                |           time(RG_v)           |
+        |<--Attestation Result------------|               |
+        |   {H,time(RX_v)}                |               |
         |                |                |               |
-     time(HR'_a)<--------+---------time(HR'_v)---->time(HR'_r)
+     time(HR'_a)<--------+-----------time(HR'_v)---->time(HR'_r)
         |                |                |               |
      time(RR_a)          |                |               |
-        |--RR{H',R{H,time(RX_v)}}--------------------->time(RA_r)
+        |---Attestation Result---------------------->time(RA_r)
+        |   {H',R{H,time(RX_v)}}          |               |
         |                |                |               |
         ~                ~                ~               ~
         |                |                |               |
+        |                |                |          time(OP_r)
 ~~~~
 {: #fig-handle-passport title="Handle-based Passport Model"}
 
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1341,7 +1341,6 @@ or might be defined relative to some other timestamp or timeticks counter.
 | VG | Value generated             | A value to appear in a Claim was created.  In some cases, a value may have technically existed before an Attester became aware of it but the Attester might have no idea how long it has had that value.  In such a case, the Value created time is the time at which the Claim containing the copy of the value was created.
 | NS | Nonce sent                  | A nonce not predictable to an Attester (recentness & uniqueness) is sent to an Attester.
 | NR | Nonce relayed               | A nonce is relayed to an Attester by another entity.
-| HT | Handle transmitted          | A new handle is transmitted by a Handle Distributor to instruct the receiving entities to transition to a new epoch.
 | HR | Handle received             | A handle is successfully received and processed by an entity.
 | EG | Evidence generation         | An Attester creates Evidence from collected Claims.
 | ER | Evidence relayed            | A Relying Party relays Evidence to a Verifier.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1511,8 +1511,6 @@ relays the Attestation Result obtained during epoch `E` (associated with handle
 
 In the illustrated scenario, the handle for relaying an Attestation Result to the Relying Party is current, while a previous handle was used to generate Verifier evaluated evidence. 
 This indicates that at least one epoch transition has occurred, and the Attestation Results may only be as fresh as the previous epoch.
-which can be taken into account in Appraisal Policies for Attestation Results
-about freshness.
 
 ~~~~
                   .-------------.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1525,17 +1525,17 @@ This indicates that at least one epoch transition has occurred, and the Attestat
      time(HR_a)<---------+-----------time(HR_v)----->time(HR_r)
         |                |                |               |
      time(EG_a)          |                |               |
-        |---Evidence{H,time(VG_a)}------->|               |
+        |---Evidence{H,time(EG_a)-time(VG_a)}----->|               |
         |                |                |               |
         |                |           time(RG_v)           |
         |<--Attestation Result------------|               |
-        |   {H,time(RX_v)}                |               |
+        |   {H,time(RX_v)-time(RG_v)}                |               |
         |                |                |               |
      time(HR'_a)<--------+-----------time(HR'_v)---->time(HR'_r)
         |                |                |               |
      time(RR_a)          |                |               |
         |---Attestation Result---------------------->time(RA_r)
-        |   {H',R{H,time(RX_v)}}          |               |
+        |   {H',R{H,time(RX_v)-time(RG_v)}}          |               |
         |                |                |               |
         ~                ~                ~               ~
         |                |                |               |
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1508,6 +1508,7 @@ After the transmission of handle `H'` a new epoch `E'` is
 established when `H'` is received by each protocol participant.  The Attester
 relays the Attestation Result obtained during epoch `E` (associated with handle
 `H`) to the Relying Party using the handle for the current epoch `H'`.
+If the Relying Party had not yet received `H'`, then the Attestation Result would be rejected, but in this example, it is received.
 
 In the illustrated scenario, the handle for relaying an Attestation Result to the Relying Party is current, while a previous handle was used to generate Verifier evaluated evidence. 
 This indicates that at least one epoch transition has occurred, and the Attestation Results may only be as fresh as the previous epoch.
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1509,9 +1509,8 @@ established when `H'` is received by each protocol participant.  The Attester
 relays the Attestation Result obtained during epoch `E` (associated with handle
 `H`) to the Relying Party using the handle for the current epoch `H'`.
 
-In the illustrated scenario, the Handle for relaying an Attestation Result to
-the Relying Party is fresh, while the handle that was used to generate the
-handle is stale. This indicates that at least one epoch has passed in between,
+In the illustrated scenario, the handle for relaying an Attestation Result to the Relying Party is current, while a previous handle was used to generate Verifier evaluated evidence. 
+This indicates that at least one epoch transition has occurred, and the Attestation Results may only be as fresh as the previous epoch.
 which can be taken into account in Appraisal Policies for Attestation Results
 about freshness.
 
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1501,7 +1501,7 @@ The Attester generates Evidence incorporating handle `H` and conveys it to the
 Verifier.
 
 The Verifier appraises that the received handle `H` is "fresh" according to the
-definition provided in {{freshness}}, and generates an Attestation Result.  The
+definition provided in {{epochfreshness}} whereby retries are required in the case of mismatching handles, and generates an Attestation Result.  The
 Attestation Result is conveyed to the Attester.
 
 After the transmission of handle `H'` a new epoch `E'` is


thomas-fossati:
This is a draft PR.

More prose is needed around the sequence diagram.

henkbirkholz:
fixed the required event tables entries

The new sequence diagram illustrate a more concise "real life" story now and st the same time intends to fixes some ambiguities that snuck in when augmenting the overall diagram layout.

Now ready for first walk-through in editors meeting.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
-As handles are distributed over the network, all involved entities receive a fresh handle at roughly the same time.
-Due to distribution over the network, there is some jitter with respect to the time the Handle is received, time(HR), for each involved entity.
-To compensate for this jitter, there is a small period of overlap (a specified offset) in which both a current handle and corresponding former handle are valid in Evidence appraisal: `validity-duration = time(HR'_v) + offset - time(HR_v)`. The offset is typically based on a network's round trip time.
-Analogously, the generation of valid Evidence is only possible, if the age of the handle used is lower than the validity-duration: `time(HR_v) - time(EG_a) < validity-duration`.
+The distribution of a new handle determines the transition from one epoch to
+the next.  The transition between epochs goes through an "overlap" phase in
+which the previous epoch is still extant while the current epoch is
+established.  The span of the overlap phase is determined solely by the
+transmission delay of the underlying communication channel,
+`time(HR)-time(HT)`.  Once the overlap phase is over, an epoch is fully
```
dthaler, line 1441: I don't think this is correct.  The overlap is `time(HR')-time(HR)`.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+established.
 
-From the point of view of a Verifier, the generation of valid Evidence is only possible, if the age of the handle used in the Evidence generation is younger than the duration of the distribution interval -- "(time(HR'_v)-time(HR_v)) - (time(HR_a)-time(EG_a)) \< validity-duration".
+In order to determine freshness of Evidence or Attestation Results, consumers
+of messages based on handles need to understand whether an epoch is fully
+established or in overlap phase.
```
dthaler, line 1446: This is confusing.  Remove it?

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
 
-Due to the validity-duration of handles, multiple different pieces of Evidence can be generated based on the same handle.
-The resulting granularity (time resolution) of Evidence freshness is typically lower than the resolution of clock-based tickcounters.
+The Handle Distributor, hosted in a dedicated entity or co-located with other
+RATS roles, is trusted by the protocol participants to behave correctly with
```
dthaler, line 1449:
```suggestion
The Handle Distributor, hosted in a dedicated entity or co-located with another
RATS role, is trusted by the protocol participants to behave correctly with
```

ericvoit:
looks good to me.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+A third option to establish the epoch of Evidence or Attestation Result
+generation - next to nonces and timestamps ({{freshness}}) - is by using
+"handles".
 
-Handles are different from nonces as they can be used more than once and can be used by more than one entity at the same time.
-Handles are different from timestamps as they do not have to convey information about a point in time, but their reception creates that information.
-The reception of a handle is similar to the event that increments a relative tickcounter.
-Receipt of a new handle invalidates a previously received handle.
+Handles are opaque, non-predictable data regularly broadcast to protocol
+participants by a Handle Distributor.  Each handle represents a specific epoch,
```
mcr, line 1433:
```suggestion
Handles are opaque, non-predictable data regularly broadcast by a Handle Distributor.  
Each handle represents a specific epoch,
```

draft-ietf-rats-architecture.md
```diff
@@ -1458,27 +1486,23 @@ The following example illustrates a hypothetical Background-Check Model solution
...
-        ~                                                 ~
-        |                                                 |
-        |                                              time(OP_r)
-        |                                                 |
+        |<----------------R{H,time(RX_v)}-|               |
+        |                |                |               |
+     time(HR'_a)<---time(HT')----------time(HR'_v)---->time(HR'_r)
+        |                |                |               |
+     time(RR_a)          |                |               |
+        |--RR{H',R{H,time(RX_v)}}--------------------->time(RA_r)
```
mcr, line 1500:
```suggestion
        |     Attestation Result
        |-------{H',R{H,time(RX_v)}}------------------>time(RA_r)
```

henkbirkholz:
```suggestion
        |---R{H',R{H,time(RX_v)}}--------------------->time(RA_r)
```

We need a legend. 'R' is short for Attestation Result.

The point of the illustration is that a Conceptual message with a stale handle is conveyed and that the Relying Party can detect that the Handle used in the Conceptual Message -- the relayed Attestation Result -- is not the current handle anymore.

Instead of adding Attestation Result text in the diagram, I simply removed the RR --> which renders the Relayed Attestation Result message a Attestation Message. This probably could be good enough, if the English text than captures that context.

mcr: I think that you should leave the text in the diagram fully expanded until we can agree that the new HT math is correct.
Then, let's see if we can find better shortcuts/abbreviations.

draft-ietf-rats-architecture.md
thomas-fossati: Can you please explain where is the confusion?

We think that being able to distinguish between a fully established epoch and one that is still in overlap phase is crucial for a receiver that needs to decide whether to accept or refuse an incoming message.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
 
-Handles are a third option to establish time-keeping next to nonces or timestamps.
-Handles are opaque data intended to be available to all RATS roles that interact with each other, such as the Attester or Verifier, in specified intervals.
-To enable this availability, handles are distributed centrally by the Handle Distributor role over the network.
-As any other role, the Handle Distributor role can be taken on by a dedicated entity or collapsed with other roles, such as a Verifier.
-The use of handles can compensate for a lack of clocks or other sources of time on entities taking on RATS roles.
-The only entity that requires access to a source of time is the entity taking on the role of Handle Distributor.
+A third option to establish the epoch of Evidence or Attestation Result
+generation - next to nonces and timestamps ({{freshness}}) - is by using
+"handles".
```
dthaler, line 1430:
```suggestion
The following example illustrates a hypothetical Passport Model
solution that uses handles instead of timestamps or nonces.
```
For consistency with the start of examples 1 and 2

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
 
-Handles are different from nonces as they can be used more than once and can be used by more than one entity at the same time.
-Handles are different from timestamps as they do not have to convey information about a point in time, but their reception creates that information.
-The reception of a handle is similar to the event that increments a relative tickcounter.
-Receipt of a new handle invalidates a previously received handle.
+Handles are opaque, non-predictable data regularly broadcast by a Handle Distributor.  
+Each handle represents a specific epoch,
+in which generated Evidence can be asserted to be fresh.
```
dthaler, line 1434: The freshness section still calls them nonces in lines 1109-1114. We should be consistent, maybe by calling them handles there.   And these two sentences are not about the example per se, they're about the definition of handles and so I think they should move to where line 1109-1114 are.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
-As handles are distributed over the network, all involved entities receive a fresh handle at roughly the same time.
-Due to distribution over the network, there is some jitter with respect to the time the Handle is received, time(HR), for each involved entity.
-To compensate for this jitter, there is a small period of overlap (a specified offset) in which both a current handle and corresponding former handle are valid in Evidence appraisal: `validity-duration = time(HR'_v) + offset - time(HR_v)`. The offset is typically based on a network's round trip time.
-Analogously, the generation of valid Evidence is only possible, if the age of the handle used is lower than the validity-duration: `time(HR_v) - time(EG_a) < validity-duration`.
+The distribution of a new handle determines the transition from one epoch to
+the next.  The transition between epochs goes through an "overlap" phase in
+which the previous epoch is still extant while the current epoch is
+established.  The span of the overlap phase is determined solely by the
+transmission delay of the underlying communication channel,
+`time(HR)-time(HT)`.  Once the overlap phase is over, an epoch is fully
```
dthaler, line 1441: I don't think this is true, since the receiver has no idea when time(HT) is.   To avoid race conditions, receivers (Verifiers / Relying Parties) need to accept both the previous handle and the current handle.  In that case, the original text inequalities are correct and this one is not.

draft-ietf-rats-architecture.md
dthaler: I don't understand your point.   If it's in overlap phase, it should still be accepted.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+RATS roles, is trusted by the protocol participants to behave correctly with
+regards to its duty of handle distribution.
 
-The following example illustrates a hypothetical Background-Check Model solution that uses handles and requires a trustworthy time source available to the Handle Distributor role.
+Consumers of Conceptual Messages based on handles must be able to safely trace
+each handle's origin back to the broadcasting Handle Distributor either using
+topological considerations (e.g., co-location), or by cryptographic means.
+
+The example in {{fig-handle-passport}} illustrates a Handle-based interaction
+using the passport topology model.
```
dthaler, line 1457: This is either redundant with the first sentence of this section, or everything above here should be moved to the freshness section.

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+Attestation Result is conveyed to the Attester.
+
+After the transmission of handle `H'` at `time(HT')` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
+
+In the illustrated scenario, the Handle for relaying an Attestation Result to
+the Relying Party is fresh, while the handle that was used to generate the
+handle is stale. This indicates that at least one epoch has passed in between,
```
dthaler, line 1476: "handle used to generate the handle" is confusing since no such concept was introduced earlier

draft-ietf-rats-architecture.md
```diff
@@ -1458,27 +1486,23 @@ The following example illustrates a hypothetical Background-Check Model solution
         |                |                |               |
         ~                ~                ~               ~
         |                |                |               |
-     time(HR_a)<---------+-------------time(HR_v)------>time(HR_r)
+     time(HR_a)<----time(HT)-----------time(HR_v)----->time(HR_r)
         |                |                |               |
      time(EG_a)          |                |               |
-        |----Evidence{time(EG_a)}-------->|               |
-        | {Handle1,time(EG_a)-time(VG_a)}|                |
+        |--E{H,time(VG_a)}--------------->|               |
```
dthaler, line 1492: Problems:
1) the term E is not defined, only Evidence
2) You cannot convey time(VG_a) without synchronized clocks
3) time(EG_a) is left in line 1491 but then not used

draft-ietf-rats-architecture.md
```diff
@@ -1458,27 +1486,23 @@ The following example illustrates a hypothetical Background-Check Model solution
...
-     time(HR_a')<--------'---------------------------->time(HR_r')
-        |                                                 |
-     time(RR_a)                                           /
-        |--Attestation Result{time(RX_v)-time(RG_v)}-->time(RA_r)
-        |    {Handle2, time(RR_a)-time(EG_a)}             |
-        ~                                                 ~
-        |                                                 |
-        |                                              time(OP_r)
-        |                                                 |
+        |<----------------R{H,time(RX_v)}-|               |
```
dthaler, line 1495: 1) R is undefined
2) time(RX_v) is not useful to convey without synchronized clocks.  There's nothing the RP can do with it in this section.

draft-ietf-rats-architecture.md
```diff
@@ -1458,27 +1486,23 @@ The following example illustrates a hypothetical Background-Check Model solution
         |                |                |               |
         ~                ~                ~               ~
         |                |                |               |
-     time(HR_a)<---------+-------------time(HR_v)------>time(HR_r)
+     time(HR_a)<----time(HT)-----------time(HR_v)----->time(HR_r)
```
dthaler, line 1489: time(HT) is not used in any computation or inequality so should be removed for consistency

dthaler (changes requested):

dthaler:
Personally, I find the new text much less clear than the old text

draft-ietf-rats-architecture.md
thomas-fossati: Sure.  However, in order to be able to _reject_ a message coming from the previous epoch a receiver needs to be able to say: _"overlap is over and the new epoch is now fully established, therefore I will assume whatever comes from a previous epoch is a replay"_

draft-ietf-rats-architecture.md
thomas-fossati: > I don't think this is true, since the receiver has no idea when time(HT) is. 

I see your point, but I wouldn't say it is a false statement.  It is however both imprecise (it should be the worst case delay among all receivers) and not really useful from an operational point of view.  Therefore I'm happy to remove it from the sentence which has clear enough prose to express the idea already.

I think we also need to highlight that the extent of the overlap, the thing previously known as "offset" (would "pad" be a better name?), is a central protocol parameter which needs calibration and agreement among all the participants to be meaningful.  It's not a unilateral decision that a protocol participant makes up by itself.

And another related thing (which I just realised, sorry for overloading the comment) is that the discussion seems to be assuming a network topology where V is _closer_ to the Handle Distributor than V - and therefore V’s epoch starts _before_ A’s epoch.  We should also consider the dual situation and therefore the case where Evidence arrives at V with a handle that V has not yet received.  This calls for extending the “offset” on the left to buffer _maybe valid_ handles?

> [...] the original text inequalities are correct and this one is not

One of the troubles I was having with the original text is that I really don’t understand what the inequality is supposed to say.  In particular, how is `time(HR_v)-time(EG_a)` a definition of “age of the handle used”?  Besides, it seems that there is an unstated assumption here that A and V have synchronised clocks - otherwise time(EG_a) and time(HR_v) are not on the same timeline and can't be meaningfully combined.

draft-ietf-rats-architecture.md
thomas-fossati: I think it should say "the handle [...] used to generate the Evidence is stale" instead?

draft-ietf-rats-architecture.md
henkbirkholz: Accepting an incoming Conceptual Message without overlap phase awareness should be very strongly discouraged, as it can break the whole RATS concept. I understand that it is not obvious or easy for a receiver to infer that it is in an overlap phase. Nevertheless, delimiting factors should be illustrated in this context as best as we can.

draft-ietf-rats-architecture.md
henkbirkholz: Dave took on the task to generate text in the Freshness section that introduces the Handle concept in the body of the document.

draft-ietf-rats-architecture.md
henkbirkholz: It was pointed out to me that my comment was mislead. Unfortunately, I have to agree.

What I intended to convey was: yes, of course the message is accepted, but blindly accepting it without awareness of an overlap phase poses a security risk in the following appraisal procedures due to lack of relevant context information. IHTH

draft-ietf-rats-architecture.md
henkbirkholz: Why couldn't it be collapsed with more than one other RATS role?

draft-ietf-rats-architecture.md
henkbirkholz, line 1476:
```suggestion
Evidence is stale. This indicates that at least one epoch has passed in between,
```
That's a typo. Evidence is correct.

draft-ietf-rats-architecture.md
henkbirkholz: I think the point was to provide more context by using a defined event in the text. I am rather cautious to agree with the notion that "used in a computation or inequality" is a needed requirement here. Also, it is used in a comparison/computation in the text.

We all agree that, for example, illustrating time(HR_a) in the diagram - although it cannot be captured by an Attester with no clock - is okay. Consequently, highlighting that time(HT) and time(HR_a) are related to effectively existing overlap phases although the delta is not easy to capture seems fine to me.

draft-ietf-rats-architecture.md
henkbirkholz: 1.) As far as I understand it, this seems to be a proposal simplify the layout of the diagram.

The current state evolved into a representation that was neither readable nor compressible:
  
```
     time(EG_a)
        |----Evidence{time(EG_a)}
        | {Handle1,time(EG_a)-time(VG_a)}
```
Whatever the solution in the end, this has to be fixed, I think (the { vs ( syntax is never introduced and not intuitive).

2. true, we have to walk through the sequence again together, I think

3. It still happens. Is it okay to not illustrates what happens, but is not conveyed in a Conceptual Message? I have some doubts that "used" only is restricted to conveyance or am I missing something obvious here?

draft-ietf-rats-architecture.md
henkbirkholz: First of all, you intended to write "where V is _closer_ to the Handle Distributor than A", I think.

Additionally, I think we came to the conclusion that it is preferable to use an event (defined in the table) in the text of the Appendix in order to provide better context. Simply removing the sentence with time(HT) at the end would go against that. Hence, I would recommend to not simply remove it, but to replace it with a paragraph that highlights why event types, such as time(HT), that can frame the extent of an overlap phase are vital to readers of the document here.

In general, Handle Distribution is in support of various scenarios. In some of them the receivers have no clocks at all, in some of them only a few - e.g. the Verifier. Maybe it would help with the scenario description, if we add a Clock to the Verfier? (so that only the swarm of Attesters has no clock)

thomas-fossati: > First of all, you intended to write "where V is _closer_ to the Handle Distributor than A", I think.

Your thinking is correct, fixed.  Thanks for spotting it!

> Additionally, I think we came to the conclusion that it is preferable to use an event (defined in the table) in the text of the Appendix in order to provide better context. Simply removing the sentence with time(HT) at the end would go against that. Hence, I would recommend to not simply remove it, but to replace it with a paragraph that highlights why event types, such as time(HT), that can frame the extent of an overlap phase are vital to readers of the document here.

I agree.  Frankly, I care more about clarity at the conceptual level rather than anything else.  So, whatever device helps to better grasp the ideas is very welcome.

> In general, Handle Distribution is in support of various scenarios. In some of them the receivers have no clocks at all, in some of them only a few - e.g. the Verifier. Maybe it would help with the scenario description, if we add a Clock to the Verfier? (so that only the swarm of Attesters has no clock)

What I was trying to say is just that the second inequality makes no sense because the two `time()` functions apply to different timelines (and are therefore the comparison is between pears and apples!) unless the underlying clocks at V and A are synchronised.  If we cannot assume the presence of clocks at nodes implementing the A role, let alone synchronised clocks, well that's another good reason for dropping the inequality, isn't it?

If instead there is at least some assumption on clock presence in some of the roles (and possibly a synchronisation requirement?), that should be clearly stated.

draft-ietf-rats-architecture.md
nedmsmith: "... co-located with other RATS roles, is trusted..."

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+definition provided above, and generates an Attestation Result.  The
+Attestation Result is conveyed to the Attester.
+
+After the transmission of handle `H'` at `time(HT')` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
+
+In the illustrated scenario, the Handle for relaying an Attestation Result to
+the Relying Party is fresh, while the handle that was used to generate the
```
nedmsmith, line 1475: The word 'while' can be removed or changed to "but" or "however"

draft-ietf-rats-architecture.md
nedmsmith: I agree there should be a legend. 
However, these examples are very complex and very terse for an architecture document. It would be better to IMO to number each message in the sequence diagram flows then explain them in more readable detail in a numbered prose section below the diagram.

draft-ietf-rats-architecture.md
dthaler: Recommend rebasing on top of PR #257

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
 
-The following example illustrates a hypothetical Background-Check Model solution that uses handles and requires a trustworthy time source available to the Handle Distributor role.
+Consumers of Conceptual Messages based on handles must be able to safely trace
+each handle's origin back to the broadcasting Handle Distributor either using
+topological considerations (e.g., co-location), or by cryptographic means.
+
+The example in {{fig-handle-passport}} illustrates a Handle-based interaction
+using the passport topology model.
+
+At `time(HT)`, the Handle Distributor broadcasts handle `H` which starts a new
```
dthaler, line 1459: It actually doesn't matter whether it's "broadcast" or only unicast to two specific entities (e.g., an Attester and Verifier) that use it.
```suggestion
At `time(HT)`, the Handle Distributor sends handle `H` which starts a new
```

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+epoch `E` for a protocol participant upon reception at `time(HR)`.
```
dthaler, line 1460: `E` in line 1492 is Evidence whereas here it's Epoch, which is confusing.   I'd like to fix line 1492 to not abbreviate Evidence

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+
+The Attester generates Evidence incorporating handle `H` and conveys it to the
+Verifier.
+
+The Verifier appraises that the received handle `H` is "fresh" according to the
+definition provided above, and generates an Attestation Result.  The
```
dthaler, line 1466: which line is "definition provided above" referring to?

draft-ietf-rats-architecture.md
```diff
@@ -1424,30 +1425,57 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+Attestation Result is conveyed to the Attester.
+
+After the transmission of handle `H'` at `time(HT')` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
```
dthaler, line 1472: and what happens if the Relying Party hasn't gotten H' yet?   (I added a brief discussion of this in PR 257, so would be good to clarify here.  This section just has an example so any answer may be ok.)

draft-ietf-rats-architecture.md
dthaler: I agree that illustrating time(HR_a) in the diagram is okay so long as it's used in a message or computation/inequality.

draft-ietf-rats-architecture.md
dthaler: In the snippet in Henk's comment there's two {} blocks on the same message, which should never happen.
{} means "includes (inside)".  That is Evidence{time(EG_a)} means that time(EG_a) is inside the signed Evidence.   I completely agree it would be good to explain the notation before or in the first example where it's used.

dthaler (changes requested):

draft-ietf-rats-architecture.md
```diff
@@ -1492,20 +1491,29 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
 
-The following example illustrates a hypothetical Passport Model
-solution that uses handles instead of nonces or timestamps.
-In this example, Evidence generation based on received handles always uses the current (most recent) handle.
-As handles are distributed over the network, all involved entities receive a fresh handle at roughly the same time.
-Due to distribution over the network, there is some jitter with respect to the time the Handle is received, time(HR), for each involved entity.
-To compensate for this jitter, there is a small period of overlap (a specified offset) in which both a current handle and corresponding former handle are valid in Evidence appraisal: `validity-duration = time(HR'_v) + offset - time(HR_v)`. The offset is typically based on a network's round trip time.
-Analogously, the generation of valid Evidence is only possible, if the age of the handle used is lower than the validity-duration: `time(HR_v) - time(EG_a) < validity-duration`.
+The example in {{fig-handle-passport}} illustrates a Handle-based interaction
+using the passport topology model.
```
dthaler, line 1495:
```suggestion
The example in {{fig-handle-passport}} illustrates a hypothetical Passport Model
solution that uses handles instead of nonces or timestamps.
```
Keep wording consistent with intros to the previous sections.

draft-ietf-rats-architecture.md
```diff
@@ -1492,20 +1491,29 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+epoch `E` for a protocol participant upon reception at `time(HR)`.
 
-From the point of view of a Verifier, the generation of valid Evidence is only possible, if the age of the handle used in the Evidence generation is younger than the duration of the distribution interval -- "(time(HR'_v)-time(HR_v)) - (time(HR_a)-time(EG_a)) \< validity-duration".
+The Attester generates Evidence incorporating handle `H` and conveys it to the
+Verifier.
 
-Due to the validity-duration of handles, multiple different pieces of Evidence can be generated based on the same handle.
-The resulting granularity (time resolution) of Evidence freshness is typically lower than the resolution of clock-based tickcounters.
+The Verifier appraises that the received handle `H` is "fresh" according to the
+definition provided in {{freshness}}, and generates an Attestation Result.  The
```
dthaler, line 1504: I'd recommend referencing the "Implicit Timekeeping using Epoch Handles" subsection rather than the full "Freshness" section.  And "according to the definition provided" there is ambiguous since it mentions that any solution has to take one of three approaches, so need to clarify which approach is shown in this example.

I.e., quoting text there:
> An actual solution also needs to take into account race conditions
> when transitioning to a new epoch, such as by using a counter signed
> by the Handle Distributor as the handle, or by including both the current and
> previous handles in messages and/or checks, by requiring retries
> in case of mismatching handles, or by buffering incoming messages
> that might be associated with a handle that the receiver has not yet
> obtained.

draft-ietf-rats-architecture.md
```diff
@@ -1516,27 +1524,26 @@ The following example illustrates a hypothetical Background-Check Model solution
...
-     time(HR_a')<--------'---------------------------->time(HR_r')
-        |                                                 |
-     time(RR_a)                                           /
-        |--Attestation Result{time(RX_v)-time(RG_v)}-->time(RA_r)
-        |    {Handle2, time(RR_a)-time(EG_a)}             |
-        ~                                                 ~
-        |                                                 |
-        |                                              time(OP_r)
-        |                                                 |
+        |---Evidence{H,time(VG_a)}------->|               |
```
dthaler, line 1530: time(VG_a) makes no sense in this example since clocks are not synchronized so the actual value is meaningless, and there's no other time(foo_a) transmitted to the attester to use as a delta.
The old (red) text had time(EG_a)-time(VG_a) which did make sense, as the delta is meaningful to the receiver (it just requires a clock in the Attester that need not be synchronized).

That's why the old text had time(RX_v)-time(RG_v) is is meaningful.

draft-ietf-rats-architecture.md
```diff
@@ -1516,27 +1524,26 @@ The following example illustrates a hypothetical Background-Check Model solution
...
+        |                |                |               |
+        |                |           time(RG_v)           |
+        |<--Attestation Result------------|               |
+        |   {H,time(RX_v)}                |               |
```
dthaler, line 1534: time(RX_v) is useless in this message for the same reason as in my previous comment.

dthaler (changes requested):

draft-ietf-rats-architecture.md
```diff
@@ -1492,20 +1491,29 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
 
-The following example illustrates a hypothetical Background-Check Model solution that uses handles and requires a trustworthy time source available to the Handle Distributor role.
+After the transmission of handle `H'` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
+
+In the illustrated scenario, the Handle for relaying an Attestation Result to
+the Relying Party is fresh, while the handle that was used to generate the
+handle is stale. This indicates that at least one epoch has passed in between,
```
mcr, line 1514:
```suggestion
In the illustrated scenario, the handle for relaying an Attestation Result to the Relying Party is current, while a previous handle was used to generate Verifier evaluated evidence. 
This indicates that at least one epoch transition has occurred, and the Attestation Results may only be as fresh as the previous epoch.
```

draft-ietf-rats-architecture.md
```diff
@@ -1492,20 +1491,28 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
-The following example illustrates a hypothetical Background-Check Model solution that uses handles and requires a trustworthy time source available to the Handle Distributor role.
+After the transmission of handle `H'` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
+
+In the illustrated scenario, the handle for relaying an Attestation Result to the Relying Party is current, while a previous handle was used to generate Verifier evaluated evidence. 
+This indicates that at least one epoch transition has occurred, and the Attestation Results may only be as fresh as the previous epoch.
+which can be taken into account in Appraisal Policies for Attestation Results
+about freshness.
```
mcr, line 1515:
```suggestion

```

draft-ietf-rats-architecture.md
```diff
@@ -1492,20 +1491,26 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+The Handle Distributor broadcasts handle `H` which starts a new
+epoch `E` for a protocol participant upon reception at `time(HR)`.
+
+The Attester generates Evidence incorporating handle `H` and conveys it to the
+Verifier.
 
-Due to the validity-duration of handles, multiple different pieces of Evidence can be generated based on the same handle.
-The resulting granularity (time resolution) of Evidence freshness is typically lower than the resolution of clock-based tickcounters.
+The Verifier appraises that the received handle `H` is "fresh" according to the
+definition provided in {{freshness}}, and generates an Attestation Result.  The
```
mcr, line 1504:
```suggestion
definition provided in {{epochfreshness}}, and generates an Attestation Result.  The
```

draft-ietf-rats-architecture.md
```diff
@@ -1492,20 +1491,26 @@ in terms of `time(RX_v)-time(RG_v)`, then the Relying Party can check
...
+Attestation Result is conveyed to the Attester.
 
-The following example illustrates a hypothetical Background-Check Model solution that uses handles and requires a trustworthy time source available to the Handle Distributor role.
+After the transmission of handle `H'` a new epoch `E'` is
+established when `H'` is received by each protocol participant.  The Attester
+relays the Attestation Result obtained during epoch `E` (associated with handle
+`H`) to the Relying Party using the handle for the current epoch `H'`.
```
mcr, line 1510:
```suggestion
`H`) to the Relying Party using the handle for the current epoch `H'`.
If the Relying Party had not yet received `H'`, then the Attestation Result would be rejected, but in this example, it is received.
```

draft-ietf-rats-architecture.md
```diff
@@ -1516,27 +1521,26 @@ The following example illustrates a hypothetical Background-Check Model solution
...
-     time(HR_a')<--------'---------------------------->time(HR_r')
-        |                                                 |
-     time(RR_a)                                           /
-        |--Attestation Result{time(RX_v)-time(RG_v)}-->time(RA_r)
-        |    {Handle2, time(RR_a)-time(EG_a)}             |
-        ~                                                 ~
-        |                                                 |
-        |                                              time(OP_r)
-        |                                                 |
+        |---Evidence{H,time(VG_a)}------->|               |
```
mcr, line 1528:
```suggestion
        |---Evidence{H,time(EG_a)-time(VG_a)}----->|               |
```

draft-ietf-rats-architecture.md
```diff
@@ -1516,27 +1521,26 @@ The following example illustrates a hypothetical Background-Check Model solution
...
+        |                |                |               |
+        |                |           time(RG_v)           |
+        |<--Attestation Result------------|               |
+        |   {H,time(RX_v)}                |               |
```
mcr, line 1532:
```suggestion
        |   {H,time(RX_v)-time(RG_v)}                |               |
```

draft-ietf-rats-architecture.md
```diff
@@ -1516,27 +1521,26 @@ The following example illustrates a hypothetical Background-Check Model solution
...
+        |                |                |               |
+     time(HR'_a)<--------+-----------time(HR'_v)---->time(HR'_r)
+        |                |                |               |
+     time(RR_a)          |                |               |
+        |---Attestation Result---------------------->time(RA_r)
+        |   {H',R{H,time(RX_v)}}          |               |
```
mcr, line 1538:
```suggestion
        |   {H',R{H,time(RX_v)-time(RG_v)}}          |               |
```

mcr (commented):
fix all time to be delta

PR #253, opened 2021-01-19
wordsmith to mention tpm  and explain execution environments

mcr:


draft-ietf-rats-architecture.md
```diff
@@ -395,11 +395,10 @@ Attesting Environments then format the claims appropriately, and typically
 use key material and
 cryptographic functions, such as signing or cipher algorithms, to
 create Evidence.
-There is no limit to or requirement on the places that an Attesting Environment can exist, but they typically are in Trusted Execution Environments (TEE), embedded Secure Elements
-(eSE), and BIOS firmware.
-An execution environment may not, by default, be capable of claims collection
-for a given Target Environment.
-Execution environments that are designed to be capable of claims collection are referred to in this document as Attesting Environments.
+There is no limit to or requirement on the places that an Attesting Environment can exist, but they typically are in a Trusted Plaform Module (TPM), a Trusted Execution Environments (TEE), an embedded Secure Elements
+(eSE), or BIOS firmware.
```
dthaler, line 399: 1) grammar arity mismatch ("a Environments", "an Elements")
2) I'm not sure a TPM alone can be an Attesting Environment since by themselves cannot collect claims as far as I understand, they require another trusted component to feed them the measurements.  If my understanding is correct, a TPM is not an Attesting Environment as defined in lines 400-401, they are a piece of an Attesting Environment.  I.e., the Attesting Environment is the TPM + whatever is feeding it measurements.

dthaler (changes requested):

draft-ietf-rats-architecture.md
mcr, line 399:
```suggestion
There is no limit to or requirement on the places that an Attesting Environment can exist, but they typically are in a  Trusted Execution Environments (TEE), an embedded Secure Elements
(eSE), and BIOS firmware.
```

PR #254, opened 2021-01-19
update what RP does for Biometric use case, fix #184

### LINKED ISSUE #184: FIDO Biometric Authentication definitions
````
   Relying Party:  Any web site, mobile application back end or service
      that does biometric authentication.
````

It sounds like the Verifier “does” the authentication, and the relying party relies on the answer?

### PR DESCRIPTION
**Pull request by mcr**

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -331,8 +331,7 @@ Attester:
 
 Relying Party:
 
-: Any web site, mobile application back end or service that does biometric authentication.
-
+: Any web site, mobile application back-end, or service that relies on digestible authentication data based on biometric information.
```
dthaler, line 334: Why is "digestable" needed?
```suggestion
: Any web site, mobile application back-end, or service that relies on authentication data based on biometric information.
```

dthaler (approved):

PR #255, opened 2021-01-19
text proposed by Henk to fix #179

### LINKED ISSUE #179: 3.4 Critical Infrastructure Control
Guy says: This section seems to say that the Connected IoT valve is going attest the cloud-based industrial control system before accepting an Open or Close command?
I’d guess what you mean is that the control system should check that the valve it’s about to open is the right one, and is probably sane, no?

mcr:


draft-ietf-rats-architecture.md
```diff
@@ -249,14 +249,14 @@ Relying Party:
...
 (e.g., power grid, traffic control, hazardous chemical processing, etc.)
 is connected to a network.  The organization managing such infrastructure
 needs to ensure that only authorized code and users can control such
-processes, and they are protected from malware or other threats.
-When a protocol operation can affect some critical
-system, the device attached to the critical equipment thus wants some
-assurance that the requester has not been compromised.  As such,
+processes, and that these processes are protected from unauthorized manipulation or other threats.
+When a protocol operation can affect a component of a critical
+system, the device attached to the critical equipment requires some assurances depending on the security context, including that: the requester and the triggering sensor have not been compromised, and the requesters and actors act on applicable policies,
```
dthaler, line 258: I think the notion of a "triggering sensor" is not important to this section.  It can be triggered by a human or malware or a policy change or time of day, or whatever else.   Suggest saying less, like the old text did.

dthaler (changes requested):

draft-ietf-rats-architecture.md
mcr, line 258:
```suggestion
system, the device attached to the critical equipment requires some assurances depending on the security context, including that: the requesting device or application have not been compromised, and the requesters and actors act on applicable policies,
```

PR #260, opened 2021-01-26
rewrite list of three failures, close #195

### LINKED ISSUE #195: 5.1 Passport Model
````
There are three ways in which the process may fail. 
````

Is there a reason to enumerate the failure cases?  How can it be argued that there are only three?  What if the device fails part way through?  What if the relying party fails?   What if the relying party is unreachable?  What if something happens that wasn’t anticipated in the appraisal policy?

Put another way, what are we learning about RATS from this example?

### PR DESCRIPTION
**Pull request by mcr**

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -641,7 +641,8 @@ can then present the Attestation Result (and possibly additional Claims)
 to a Relying Party, which then compares this information against its own
 appraisal policy.
 
-There are three ways in which the process may fail.
+Three ways in which in which the process fail include:
```
mcr, line 644:
```suggestion
Three ways in which the process fail are as follows.
```

draft-ietf-rats-architecture.md
```diff
@@ -641,7 +641,8 @@ can then present the Attestation Result (and possibly additional Claims)
 to a Relying Party, which then compares this information against its own
 appraisal policy.
 
-There are three ways in which the process may fail.
+Three ways in which the process fail are as follows.
+
 First, the Verifier may not issue a positive Attestation Result due to the Evidence not passing the Appraisal Policy for Evidence.
 The second way in which the process may fail is when the Attestation Result is examined by the Relying Party, and based upon the Appraisal Policy for Attestation Results, the result does not pass the policy.
 The third way is when the Verifier is unreachable or unavailable.
```
mcr, line 648:
```suggestion
Three ways in which the process may fail include: 

* First, the Verifier may not issue a positive Attestation Result due to the Evidence not passing the Appraisal Policy for Evidence.

* The second way in which the process may fail is when the Attestation Result is examined by the Relying Party, and based upon the Appraisal Policy for Attestation Results, the result does not pass the policy.

* The third way is when the Verifier is unreachable or unavailable.
```

PR #263, opened 2021-01-27
fixes #235 (slight cognitive dissonance in the "Roles and Entities" section)

### LINKED ISSUE #235: slight cognitive dissonance in the "Roles and Entities" section
This sentence:
```
   An entity in the RATS architecture includes at least one of the roles
   defined in this document.
```
creates a bit of cognitive dissonance.

In the Terminology section, the Endorser, Ref-val Provider, RP Owner and Verifier Owner are defined as "entities", however, they don't include any of the roles (Verifier, RP, Attester).

henkbirkholz:
harmonized phrasing in the terminology section -> roles are finally roles,
changed Terminology item format,
strayed off and fixed a capitalized "conceptual messages"


henkbirkholz:
While addressing #235 I realized that I am not really happy with:

* the use of "entity" (issue). Was there a reason why we are using such a broad term here ? The term "actor" is used in other attestation-related documents. Would that be an alternative here?

* the use of "Artifacts" (nit). Used here as a section name and never again. I can see why this spawned into existence: the Artifacts Section contains Conceptual Messages, Policies, and then the term Claims. Maybe we should split this section in three?

PR #267, opened 2021-01-29
phrasing of intro to "unprotected evidence"

### LINKED ISSUE #212: 7.4 Verifier unsigned CBOR
````
  A conveyance protocol that provides authentication and integrity
   protection can be used to convey unprotected Evidence, assuming the
  
````

I assume this is an escape hatch for unsigned CBOR.
But I’m not sure the sentence makes sense…  if the conveyance protocol provides protection, then it’s not unprotected evidence?
Or is there some other point we’re trying to get at?

### PR DESCRIPTION
**Pull request by henkbirkholz**
Fixes #212

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -100,6 +100,7 @@ informative:
     title: "DICE Certificate Profiles"
     target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Certificate-Profiles-r01_3june2020-1.pdf
   I-D.birkholz-rats-tuda: rats-tuda
+  I-D.birkholz-rats-uccs: uccs
   I-D.ietf-teep-architecture: teep-arch
   TCGarch:
     title: "Trusted Platform Module Library - Part 1: Architecture"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by William-PanWei — APPROVED** [review]

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -898,10 +899,10 @@ some way of verifying the signature on the Evidence.  This may be in the form of
 The nature of how the Verifier manages to validate the signatures produced by the Attester is critical to the secure operation of an Attestation system, but is not the subject of standardization within this architecture.
 
 A conveyance protocol that provides authentication and integrity protection can be used
-to convey unprotected Evidence, assuming the following properties exists:
+to convey Evidence that is otherwise unprotected (e.g., not signed). Appropriate conveyance of unprotected Evidence (e.g., {{-uccs}}) relies on the following conveyance protocol's protection capabilities:
```
dthaler, line 902:
```suggestion
to convey Evidence that is otherwise unsigned. Appropriate conveyance of unsigned Evidence (e.g., {{-uccs}}) relies on the following conveyance protocol's protection capabilities:
```
Ok as is but would prefer suggestion above if the meaning is the same, for brevity

PR #268, opened 2021-02-02
Avoid "mutual attestation" term and reference DAA discussion

### LINKED ISSUE #209: 7.3 Relying Party owner --- Mutual Authentication
````
In such a case,
   _mutual authentication_ or attestation might be needed, in which case
   typically one side's Evidence must be considered safe to share with
   an untrusted entity, in order to bootstrap the sequence.
````

Does this term (_Mutual Authentication_) need to be defined?
Aren’t there lots of cases where mutual attestation might be employed?

### PR DESCRIPTION
**Pull request by dthaler**
1) Avoid the term "mutual attestation"
2) Add reference to CCC whitepaper for discussion of anonymous attestation

Fixes #209

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -123,6 +123,11 @@ informative:
     target: "https://csrc.nist.gov/glossary/term/strength_of_function"
     author:
       org: "NISC"
+  CCC-DeepDive:
+    author:
+      org: "Confidential Computing Consortium"
+    title: "Confidential Computing Deep Dive v1.0"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -123,6 +123,11 @@ informative:
     target: "https://csrc.nist.gov/glossary/term/strength_of_function"
     author:
       org: "NISC"
+  CCC-DeepDive:
+    author:
+      org: "Confidential Computing Consortium"
+    title: "Confidential Computing Deep Dive v1.0"
```
mcr, line 129:
```suggestion
    title: "Confidential Computing Deep Dive"
```

PR #270, opened 2021-02-02
capitalize Claims in appropriate places

mcr:


draft-ietf-rats-architecture.md
```diff
@@ -193,7 +193,7 @@ machines attached to their network, for purposes such as inventory,
 audit, anomaly detection, record maintenance and/or trending reports (logging).
 The network operator may also want a policy
 by which full access is only granted to devices that meet some definition
-of hygiene, and so wants to get claims about such information and verify
+of hygiene, and so wants to get Claims about such information and verify
```
dthaler, line 196: I don't feel strongly either way but it does seem odd to me to capitalize a term before it's defined in line 582, for the same reason it's not capitalized in the abstract.  (Relying Party etc are defined earlier, but Claim is only defined later.)

I am ok either way, but I expect this is an area that will confuse the RFC editor.

mcr, line 196:
```suggestion
of hygiene, and so wants to get claims about such information and verify
```

PR #273, opened 2021-02-02
Issue #215 - Attestation Results

### LINKED ISSUE #215: 8.3 Attestation Results - authorization
````
he passport model) or a Relying Party (in the background-check
   model) to indicate that the _Attester should not be treated as
   authorized_ and may be in need of remediation.  In some cases, it may
   even indicate that the Evidence itself cannot be authenticated as
   being correct.
````

Is there a different way to say this?  I assume in Passport, we’re not relying on the Attester to report itself as untrustworthy?
I think it’s more that it doesn’t get a passport, so it has nothing to hand the dude at the immigration desk?  i.e. untrustworthiness is indicated by lack of a passport?

### PR DESCRIPTION
**Pull request by nedmsmith**
Reworded 2nd and 3rd paragraphs fix #215

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -999,7 +999,7 @@ a Relying Party (in the background-check model) to indicate that the Attester
 should not be treated as authorized and may be in need of remediation.  In some cases,
 it may even indicate that the Evidence itself cannot be authenticated as being correct.
 
-By default, the Relying Party assumes the Attester does not comply with Verifier appraisals. 
+By default, the Relying Party does not believe the Attester to be compliant. 
 Upon receipt of an authentic Attestation Result and given the Appraisal Policy for Relying Party is satisfied. 
 The Attester is allowed to perform the prescribed actions or access. 
 The simplest such Appraisal Policy might authorize granting the Attester full access or control over the resources guarded by the Relying Party.  
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -994,17 +994,18 @@ Attestation Results may carry a boolean value indicating compliance or non-compl
 The quality of the Attestation Results depend upon the ability of the Verifier to evaluate the Attester.
 Different Attesters have a different _Strength of Function_ {{strengthoffunction}}, which results in the Attestation Results being qualitatively different in strength.
 
-A result that indicates non-compliance can be used by an Attester (in the passport model) or
+An Attestation Result that indicates non-compliance can be used by an Attester (in the passport model) or
 a Relying Party (in the background-check model) to indicate that the Attester
 should not be treated as authorized and may be in need of remediation.  In some cases,
 it may even indicate that the Evidence itself cannot be authenticated as being correct.
 
-An Attestation Result that indicates compliance can be used by a Relying Party to make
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -994,17 +994,18 @@ Attestation Results may carry a boolean value indicating compliance or non-compl
...
 should not be treated as authorized and may be in need of remediation.  In some cases,
 it may even indicate that the Evidence itself cannot be authenticated as being correct.
 
-An Attestation Result that indicates compliance can be used by a Relying Party to make
-authorization decisions based on the Relying Party's appraisal policy.  The simplest such
-policy might be to simply authorize any party supplying a compliant Attestation Result
-signed by a trusted Verifier.  A more complex policy might also entail comparing information
-provided in the result against Reference Values, or applying more complex logic
-on such information.
+By default, the Relying Party assumes the Attester does not comply with Verifier appraisals.
```
mcr, line 1002: This seems really awkward.

```suggestion
By default, the Relying Party does not believe the Attester to be compliant. 
```

draft-ietf-rats-architecture.md
```diff
@@ -994,17 +994,18 @@ Attestation Results may carry a boolean value indicating compliance or non-compl
...
 
-An Attestation Result that indicates compliance can be used by a Relying Party to make
-authorization decisions based on the Relying Party's appraisal policy.  The simplest such
-policy might be to simply authorize any party supplying a compliant Attestation Result
-signed by a trusted Verifier.  A more complex policy might also entail comparing information
-provided in the result against Reference Values, or applying more complex logic
-on such information.
+By default, the Relying Party assumes the Attester does not comply with Verifier appraisals. 
+Upon receipt of an authentic Attestation Result and given the Appraisal Policy for Relying Party is satisfied. 
+The Attester is allowed to perform the prescribed actions or access.
```
mcr, line 1004:
```suggestion
Upon receipt of an authentic Attestation Result and given the Appraisal Policy for Relying Party is satisfied, then
the Attester is allowed to perform the prescribed actions or access. 
```

mcr (changes requested):
some clarifications needed

draft-ietf-rats-architecture.md
dthaler: +1 to mcr's suggested text

draft-ietf-rats-architecture.md
dthaler: Better, but "Appraisal Policy for Relying Party" also needs to be "Appraisal Policy for Attestation Results" which is the defined term.

dthaler (changes requested):

draft-ietf-rats-architecture.md
William-PanWei: One more question: Checking the Attestation Result's authenticity is included in the Appraisal Policy for Attestation Results or not?
I.e., should we call an Attestation Result authentic only after the Appraisal Policy for Attestation Results is satisfied?

William-PanWei (changes requested):

draft-ietf-rats-architecture.md
mcr: okay, what wording change do you suggest?

Issue #193, opened 2020-12-01
4.4 --- Attesting Environment


mcr:
If a TPM is the Attesting environment, then in the case where this system had a TPM, this statement would not be true, no?
In other words, shouldn’t this example be restricted to attestation without a TPM?  (I think you can say something like “hardware root of trust for verification” or something if we should avoid naming technologies)

In contrast, I think in a TPM system, the firmware initializes the Attesting Environment, the TPM is the attesting environment, and the same attesting environment is used by all layers.

dthaler:
For context, Guy's comment was on the first sentence in this paragraph:

> After the boot sequence is started, the BIOS conducts the most
>    important and defining feature of layered attestation, which is that
>    the successfully measured Target Environment B now becomes (or
>    contains) an Attesting Environment for the next layer.  This
>    procedure in Layered Attestation is sometimes called "staging".  It
>    is important that the new Attesting Environment B not be able to
>    alter any Claims about its own Target Environment B.  This can be
>    ensured having those Claims be either signed by Attesting Environment
>    A or stored in an untamperable manner by Attesting Environment A.

I don't know how TCG defines it, but my understanding is that a TPM requires another entity to feed it measurements which the TPM measures, the TPM doesn't actively collect claims itself.  Thus to me, the TPM is not the Attesting Environment.  The Attesting Environment is the _combination of the TPM + whatever is feeding it the measurements_.   Viewed that way, I believe the text is true as written.

henkbirkholz:
Yes. We need text that captures that prominently and then also in that text start with "the composite of TPM & root of trust for measurement" forms an Attesting Environment instead of "he TPM is not the Attesting Environment". Starting with the latter seems to produce reflexive and unnecessary knee-jerk reactions that we can also easily resolve, finally. Does that sound like a feasible approach?

dthaler:
> Yes. We need text that captures that prominently and then also in that text start with "the composite of TPM & root of trust for measurement" forms an Attesting Environment instead of "he TPM is not the Attesting Environment". Starting with the latter seems to produce reflexive and unnecessary knee-jerk reactions that we can also easily resolve, finally. Does that sound like a feasible approach?

Yes I like that.

mcr:
section: ## Two Types of Environments of an Attester
now includes text:

~~~~
There is no limit to or requirement on the types of hardware or software environments that can be used to implement an Attesting Environment, for example: Trusted Execution Environments (TEEs), embedded Secure Elements
[...]

PR #275, opened 2021-02-02
Issue #189 - Claims Collection

### LINKED ISSUE #189: 4.3 claims collection
````
An execution environment
   may not, by default, be capable of claims collection for a given
````

Why not?  What’s the message you want to convey here?  What would disqualify an execution environment from being capable of claims collection?
How would you know if an Execution Environment is an Attesting Environment?

nedmsmith:
Added a sentence after "An execution environment may not, by default, be capable of claims collection..."
fix #189


mcr (approved):

dthaler (commented):
PR #282 solves this in a different way which also addresses issue #193.
I would assert that merging 282 would obviate the need for any other change and so this PR wouldn't be needed.

mcr:
closing, as other text fixed this well enough.

PR #278, opened 2021-02-04
Integrity at different layers

mcr:


dthaler (approved):
Ok but this PR merges into the `capitalize-claims` branch instead of master.
Recommend rebasing on master so that order of merge doesn't matter.

William-PanWei (approved):

mcr:
> Ok but this PR merges into the `capitalize-claims` branch instead of master.
> Recommend rebasing on master so that order of merge doesn't matter.

rebased.

PR #287, opened 2021-02-05
More fixes to handle example

### LINKED ISSUE #198: Handles appendix needs some care

dthaler:
Biggest bug in the last pull request is that the diagram had the attester sending a different Attestation Result than the one the Verifier gave it, which would invalidate the signature.

Second issue was that the lines didn't line up.

Third issue was that it was missing the tie to the "epoch window" discussion in {{epochfreshness}} to explain how the Relying Party deals with the mismatched handles that the example diagram has.

Fixes #198


mcr (approved):
Thanks for aligning the diagram, I knew I had forgotten something.
The changes look good to me.

thomas-fossati (approved):
Ship it!

PR #288, opened 2021-02-09
added sentence to say why attestation results are relevant

### LINKED ISSUE #203: 5.1 - password model, evaluation of evidence
````
The format of the Evidence on the other hand is only
   constrained by the Attester-Verifier remote attestation protocol.
````

Yes but for real passports, the Relying Party must separately evaluate the trustworthiness of 192 different Verifiers, i.e., nation states, no?  This seems to imply that relying parties had better get ready to validate a limitless variety of Attestation Results?

[mcr: yes, in fact, in real passports, Relying Parties do not treat all 192 nation states equally, which is why, I'm told, Israeli Mossad agents apprently like Canadian passports...]

### PR DESCRIPTION
**Pull request by mcr**
fix #203

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -691,7 +691,7 @@ includes an Attestation Result, in this model the details of that protocol
 constrain the serialization format of the Attestation Result. The
 format of the Evidence on the other hand is only constrained by the
 Attester-Verifier remote attestation protocol.
-This implies that interoperability and standardization is more relevant for Attestation Results than it is for Evidence (in this model).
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -691,6 +691,7 @@ includes an Attestation Result, in this model the details of that protocol
 constrain the serialization format of the Attestation Result. The
 format of the Evidence on the other hand is only constrained by the
 Attester-Verifier remote attestation protocol.
+This implies that interoperability and standardization is more relevant for Attestation Results than it is for Evidence (in this model).
```
mcr, line 694:
```suggestion
This implies that interoperability and standardization is more relevant for Attestation Results than it is for Evidence.
```

PR #290, opened 2021-02-27
Fix layered attestation diagram

dthaler:
Once we added Reference Value Provider as a role, the diagram wasn't updated.  The "A, B, and C" part is about reference values, not the endorsing of signing capabilities, which is just about A.


mcr (approved):

William-PanWei (approved):

layered-attester.txt
```diff
@@ -1,8 +1,12 @@
-    .----------.                    .----------.
-    |          |                    |          |
-    | Endorser |------------------->| Verifier |
-    |          |    Endorsements    |          |
-    '----------'  for A, B, and C   '----------'
+    .-------------.   Endorsement for A
```
nedmsmith, line 1: I think it is correct to include "Endorsement for A, B and C" since it is possible that attestation keys could exist at other layers besides the first (A). The definition of endorsement focuses on the "signing capability" which doesn't have to be centralized into a single part of the device / platform.

nedmsmith (changes requested):
requesting adding ", b and c" to the endorsements portion of the figure.

layered-attester.txt
dthaler: Endorsement is defined as "A secure statement that an Endorser vouches for the integrity of an Attester's various capabilities such as Claims collection and Evidence signing."   A is what vouches for the integrity of B's capabilities and similarly B is what vouches for C.

We have no discussion anywhere of having multiple components vouch for the same target environment, so I disagree with Ned's proposal as being insufficiently motivated by any other text.  This PR is just to make the doc be internally consistent and fix an accidental oversight when RV providers were added distinct from Endorsers.

nedmsmith: I agree that the diagram requires 'fixing' from the change made earlier. 
I disagree that that attesting environment A is effectively the Endorser for environment B and so on. While it is correct that A 'vouches' for B and B for C etc... I disagree that A can vouch for the signing capability of B beyond what is captured in software. Software has to run on hardware that A isn't capable of assessing. If B and C are not hardware (only A) then I agree with Dave's analysis. If B and C can be hardware (that isn't A hardware) then it is possible for B and C to also have Endorsers.

dthaler: The text says:
> The device illustrated in {{layered}} includes (A) a BIOS stored in read-only memory,
> (B) an operating system kernel, and (C) an application or workload.

So B and C are already explicitly stated to be software per above.

mcr (approved):

Issue #293, opened 2021-03-09
Clarifying layered attestation


laurencelundblade:
This is about the text that describes layered attestation and Figure 3, the one with environments A, B and C.

I think the critical point about layered attestation (if I understand it) is that the remote Verifier must evaluate the evidence for target environment B with a positive result as a precondition to evaluating the evidence for C. This critical point isn't mentioned in the text.

The other possibility is Attesting Environment A is also the Verifier for the evidence from target environment B. There are two roles in the same box, Attester and Verifier. This seems to be implied by the paragraph after Figure 3. The paragraph that starts "Attesting Environment A... has to ensure the integrity of the boot loader (Target Environment B)". However, figure 3 only shows a remote Verifier.

A third possibility is A does ensure integrity of B, but it's not Attestation. It is just typical staged boot that is being described. It is not layered attestation. I think we should distinguish staged boot from local Attestation Verification. It doesn't seem right to claim that all staged boot is local Attestation Verification and thus RATS applies to all staged boot.

In the cases where A does locally ensure the integrity of B either via staged boot or a local Verifier, then the remote verifier can actually trust the evidence about C without first evaluating the evidence from B.

Or, all three of the above could be intended. There is nothing wrong with implementations that do any of these or combinations of them.

I think the text should decide which of the three it is going after here and be clear about them.

PenglinYang-CMCC:
Agreed，trust relations between different layers should be clarified. It seems that the concept of trust chain could be used to describe this trust and attestation relations in different layers.

dthaler:
I think the intent of the text is the first approach, i.e., the one for which Laurence says:

> I think the critical point about layered attestation (if I understand it) is that the remote Verifier must evaluate the evidence for target environment B with a positive result as a precondition to evaluating the evidence for C. This critical point isn't mentioned in the text.

PenglinYang-CMCC:
In the first approach, if A is the BIOS (root of trust ?), and B is the bootloader, I'm not sure if bootloader has the ability to communicate with remote verifier via network or other methods.

PenglinYang-CMCC:
In the first approach, if A is the BIOS (root of trust ?), and B is the bootloader, I'm not sure if bootloader has the ability to communicate with remote verifier via network or other methods.



BR~

 
From: Dave Thaler
Date: 2021-04-09 07:00
To: ietf-rats-wg/architecture
CC: PenglinYang-CMCC; Comment
[...]

PR #294, opened 2021-03-10
Reword description of layered attestation

laurencelundblade:
I've tried to stick to exactly describing the construct of layered attestation (as I understand it). I've left out discussion of terms like root of trust and trust anchor because they aren't specific to layered attestation and they are not necessary to do layered attestation.


mcr (commented):
I find the result more readable, but I'm not yet clear if there are semantic changes.

dthaler:
This PR introduces a significant change in meaning that I'm not comfortable with, so we should discuss.
Specifically, previously there was 1 attester with multiple layers.
This changes the terminology so that each layer is an attester, which is a terminology change I don't think we should make.

henkbirkholz:
This PR is defiantly a significant change to the content and would create ripple effects throughout the document. I am hesitant to open that can of worms, actually. Maybe we can isolate an issue that motivated this set of changes?

draft-ietf-rats-architecture.md
```diff
@@ -424,17 +424,19 @@ of the TPM together with whatever component is feeding it the measurements.
...
-Normally, Claims are not self-asserted, rather a previous layer acts as the Attesting Environment for the next layer.
-Claims about a root of trust typically are asserted by Endorsers.
+For layered attestation there is a root Attester and several other attesters that are not roots.
+The root attester is also called the initial Attester.
+
+The Verifier comes to trust the root Attester in any of the normal ways, for example, through an Endorsement.
+Trust in the Attesters layered above are not established in any of the normal ways.
+Instead, the root attester provides Evidence about the other attester layers. 
+
+That is, the root Attester collects and provides claims about Attester in the layer immediately above.
```
mcr, line 434: This implies that there are multiple Attesters, (one per layer), and this is change, because the original text had one Attester collecting Claims from multiple layers.

This change would have implications to many other documents, TEEP, EAT, and CCC white paper and other sections of this document.

mcr (changes requested):

laurencelundblade:
I've replace "Attester" with "Attesting Environment". It was my mistake confusing the two. Sorry about that.

I don't have an agenda to change what layered attestation is. I am just trying to clarify it and be helpful by providing new text. I have some trouble understanding the text that is there now so maybe my new text is off.

See also issue #293. The existing text talks about one layer verifying the integrity of another. Is that verification local Attestation or staged boot? Is staged boot required for layered attestation?

mcr:
Current text has evolved, and is compatible with current EAT ID text, closing.

PR #296, opened 2021-03-12
Editorial review of the whole document

henkbirkholz:
This is the portion of the complete review that contains minor/editorial changes aggregated in a single PR and various commits as discussed in the last editor meeting.

PR #327, opened 2021-03-29
Attempt to clarify slots discussion

### LINKED ISSUE #301: Some confusion about the 2nd paragraph in the Composite Device (3.5) section
> But other slots can communicate with the main slot by the links between them inside the router.

I am not certain what this sentence is trying to tell me in this context?

dthaler:
Fixes #301

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -359,7 +359,7 @@ The text below summarizes the activities conducted by the roles illustrated in {
 An Attester creates Evidence that is conveyed to a Verifier.
 
 A Verifier uses the Evidence, any Reference Values from Reference Value Providers, and any Endorsements from Endorsers,
-by applying an Appraisal Policy for Evidence to assess the trustworthiness of the Attester. 
+by applying an Appraisal Policy for Evidence to assess the trustworthiness of the Attester.
 This procedure is called the appraisal of Evidence.
 
 Subsequently, the Verifier generates Attestation Results for use by Relying Parties.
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):
lgtm

draft-ietf-rats-architecture.md
```diff
@@ -498,10 +498,10 @@ For example, a carrier-grade router consists of a chassis and multiple slots.
 The trustworthiness of the router depends on all its slots' trustworthiness.
 Each slot has an Attesting Environment such as a TEE collecting the
 Claims of its boot process, after which it generates Evidence from the Claims.
-Among these slots, only a main slot can communicate with the Verifier
-while other slots cannot. But other slots can communicate with the main
-slot by the links between them inside the router. So the main slot collects
-the Evidence of other slots, produces the final Evidence of the whole router and
+Among these slots, only a "main" slot can communicate with the Verifier
+while other slots cannot, but other slots can communicate with the main
+slot across the links between slots inside the router. The main slot thus collects
```
mcr, line 503:
```suggestion
while other slots cannot, but other slots can communicate with the "main"
slot across the links between slots inside the router. The "main" slot thus collects
```

mcr: I'm not sure why "main" is quoted, but should it be consistent?

dthaler: It was quoted the first time as being defined in context, and then used unquoted thereafter.  See for example https://blog.apastyle.org/apastyle/2013/02/quotation-mark-uses-other-than-quotes.html which quotes the guidance:
```
To introduce a word or phrase used as an ironic comment, as slang, or as an invented or coined expression. Use quotation marks the first time the word or phrase is used; thereafter, do not use quotation marks. (p. 91)
```
(Here "main" router is an invented or coined expression)

dthaler:
I see PR #296 contains a different rewrite of this text.  We should compare them and decide which one to take, or whether to take some of each.

PR #339, opened 2021-04-16
rewrite section on confidentiality of attester key material, 

### LINKED ISSUE #308: Similarity of examples in the Attestation Key Provisioning Processes (12.1.2) section
> Confidentiality protection can be realized via physical provisioning facility security involving no encryption at all. For low-security use cases, this might be simply locking doors and limiting personnel that can enter the facility. For high-security use cases, this might involve a special area of the facility accessible only to select security-trained personnel.

Both examples basically seem to be the same here?

### LINKED ISSUE #309: Some missing context in the Attestation Key Provisioning Processes (12.1.2) section
> Another way to provision key material is to generate it on the device and export the validation key. If public-key cryptography is being used, then only integrity is necessary. Confidentiality is not necessary.

A few words on the “why that is the case” are needed here.

mcr:
close #308 and close #309


draft-ietf-rats-architecture.md
```diff
@@ -1328,16 +1328,21 @@ One way to provision key material is to first generate it external to the device
...
+The cryptography that is used to enable confidentiality protection of the attestation key comes with its own requirements to be secured.
+This results in recursive problems, as the key material used to provision attestation keys must again somehow have been provisioned securely beforehand (requiring an additional level of protection, and so on).
 
-In general, a combination of some physical security measures and some cryptographic measures is used to establish confidentiality protection.
+So, this is why, in general, a combination of some physical security measures and some cryptographic measures is used to establish confidentiality protection.
 
-Another way to provision key material is to generate it on the device and export the validation key. If public-key cryptography is being used, then only integrity is necessary. Confidentiality of public keys is not necessary.
+When key material is generated within a device and the secret part of it never leaves the device,
```
dthaler, line 1341: "When" sounds like it's still referring to the above case rather than an alternative.  Need to clarify, or else create two subsections.

draft-ietf-rats-architecture.md
```diff
@@ -1328,16 +1328,21 @@ One way to provision key material is to first generate it external to the device
...
+then the problem may lessen.  For public-key cryptography, it is, by definition, not necessary to
+maintain confidentiality of the public-key: however integrity of the chain of custody of the public is necessary in order to avoid attacks where an attacker is able to enroll an attestation key that they control.
```
dthaler, line 1343:
```suggestion
maintain confidentiality of the public key: however integrity of the chain of custody of the public key is necessary in order to avoid attacks where an attacker is able to enroll an attestation key that they control.
```

draft-ietf-rats-architecture.md
dthaler, line 1341: "When" sounds like it's still referring to the above case rather than an alternative.  Need to clarify, or else create two subsections.

draft-ietf-rats-architecture.md
dthaler: "enroll" is undefined

draft-ietf-rats-architecture.md
```diff
@@ -1328,16 +1328,21 @@ One way to provision key material is to first generate it external to the device
...
 
-In all cases, attestation key provisioning must ensure that only attestation key material that is generated by a valid Endorser is established in Attesters.
-For many use cases, this will involve physical security at the facility, to prevent unauthorized devices from being manufactured that may be counterfeit or incorrectly configured.
+To summarize: attestation key provisioning must ensure that only attestation key material that is generated by a valid Endorser is established in Attesters.
```
dthaler, line 1345: "by a valid Endorser" implies that one MUST have an Endorser role, maybe that's too strong?

draft-ietf-rats-architecture.md
mcr, line 1343:
```suggestion
then the problem may lessen.  
For public-key cryptography, it is, by definition, not necessary to maintain confidentiality of the public-key: 
however integrity the public key is necessary in order to avoid attacks where an attacker is able to enroll an attestation key that they control.
```

draft-ietf-rats-architecture.md
```diff
@@ -1328,16 +1328,21 @@ One way to provision key material is to first generate it external to the device
...
+This results in recursive problems, as the key material used to provision attestation keys must again somehow have been provisioned securely beforehand (requiring an additional level of protection, and so on).
 
-In general, a combination of some physical security measures and some cryptographic measures is used to establish confidentiality protection.
+So, this is why, in general, a combination of some physical security measures and some cryptographic measures is used to establish confidentiality protection.
 
-Another way to provision key material is to generate it on the device and export the validation key. If public-key cryptography is being used, then only integrity is necessary. Confidentiality of public keys is not necessary.
+When key material is generated within a device and the secret part of it never leaves the device,
+then the problem may lessen.  For public-key cryptography, it is, by definition, not necessary to
+maintain confidentiality of the public key: however integrity of the chain of custody of the public key is necessary in order to avoid attacks where an attacker is able to enroll an attestation key that they control.
```
dthaler, line 1343: what does "enroll" mean?  undefined term.

draft-ietf-rats-architecture.md
```diff
@@ -1324,20 +1324,29 @@ Measures for a highly protected system could include specialized hardware that i
...
+So, this is why, in general, a combination of some physical security measures and some cryptographic measures is used to establish confidentiality protection.
 
-In general, a combination of some physical security measures and some cryptographic measures is used to establish confidentiality protection.
+#### On-Device Key Generation
 
-Another way to provision key material is to generate it on the device and export the validation key. If public-key cryptography is being used, then only integrity is necessary. Confidentiality of public keys is not necessary.
+When key material is generated within a device and the secret part of it never leaves the device,
+then the problem may lessen.  For public-key cryptography, it is, by definition, not necessary to
+maintain confidentiality of the public key: however integrity of the chain of custody of the public key is necessary in order to avoid attacks where an attacker is able to enroll an attestation key that they control.
```
mcr, line 1347:
```suggestion
maintain confidentiality of the public key: however integrity of the chain of custody of the public key is necessary in order to avoid attacks where an attacker is able get a key they control endorsed.
```

draft-ietf-rats-architecture.md
```diff
@@ -1324,20 +1324,29 @@ Measures for a highly protected system could include specialized hardware that i
...
 
-In all cases, attestation key provisioning must ensure that only attestation key material that is generated by a valid Endorser is established in Attesters.
-For many use cases, this will involve physical security at the facility, to prevent unauthorized devices from being manufactured that may be counterfeit or incorrectly configured.
+To summarize: attestation key provisioning must ensure that only attestation key material that is generated by a valid Endorser is established in Attesters.
```
mcr, line 1349:
```suggestion
To summarize: attestation key provisioning must ensure that only valid attestation key material is established in Attesters.
```

PR #343, opened 2021-05-19
signing does not provide confidentiality protection

thomas-fossati:


William-PanWei (approved):

henkbirkholz (approved):
Reads like I might have broken that meaning & I am in violent agreement with the fix.

Issue #344, opened 2021-05-28
Missing arrow in Figure 2?


thomas-fossati:
ISTM that the Attesting Environment in Figure 2 should also have an (optional) inbound arrow for Evidence from another Attesting Enviroment, maybe a sub-Attester.  Otherwise the arrows from Attesters B & C into the lead Attesting Environment shown in Figure 4 might come a bit as a surprise. (And the arrow labelled "Evidence for bootloader" in Figure 3 too).

nedmsmith:
I'm indifferent as to whether it needs fixing. The fact that someone raised a question suggests it isn't clear. Someone else however could look at the new content and be confused by it as well. Though, I generally error on the side of inclusion so long as the additional detail is correct and doesn't bias toward a particular implementation.

mcr:
I don't think the original diagram needs to be fixed.
It might be that a second diagram is needed somewhere else, but I think that changing this diagram will cause more confusion.

What I think we are doing is being overly mathematical, and thinking that the Architecture document is trying replace all the documents and work that has gone before, or that it is a Design Requirements document.

thomas-fossati:
> I don't think the original diagram needs to be fixed.

@laurencelundblade too, independently, got confused by the current layout in the figures -- see 2nd para of https://mailarchive.ietf.org/arch/msg/rats/UuFwOfLnRitCbnfKpE-IgZ9ANcU/

So there seems to be some kind of problem with it.

> It might be that a second diagram is needed somewhere else, but I think that changing this diagram will cause more confusion.

I am not sure multiplying the number of pics would fix what seems to be like a tiny lack of internal coherency in what we currently have.

henkbirkholz:
@thomas-fossati, if this is not resolved, could you fix that "tiny lack of internal coherency"?

thomas-fossati:
hi @henkbirkholz ! #347 already fixes said "tiny lack of internal coherency"

thomas-fossati:
> [...] what about the rest of PR, is that resolved?

PR #347 fully resolves this issue.

mcr:
Consensus to add a new diagram, at the place where the external attester is first mentioned.

thomas-fossati:
@mcr, all: I have been staring at sections 3.1 and §3.2 (and in between) for a good 15 minutes now without being able to generate a satisfactory solution.

I have decided that the document is so good that it can come out of this "tiny lack of internal coherency" unscathed.  I'm happy to withdraw the issue.

[...]

PR #347, opened 2021-06-28
Update twotypes-env.txt

nedmsmith:
Added an optional attesting environment that supplies evidence to a lead attesting env.


thomas-fossati (approved):
LGTM, thanks!

nedmsmith:
Circlecl fails to build but it builds successfully on my local machine. Someone familiar with circlecl should take a look.

thomas-fossati:
> Circlecl fails to build but it builds successfully on my local machine. Someone familiar with circlecl should take a look.

See https://github.com/martinthomson/i-d-template/issues/293

TL;DR drop CircleCI and move to GitHub Actions :-)

mcr:
But, about the proposed change to the diagram.
I don't think that this helps with understanding at this point in the document.

thomas-fossati:
> But, about the proposed change to the diagram.
> I don't think that this helps with understanding at this point in the document.

It's not made explicit In the PR, but Ned's patch addresses issue #344.  Maybe let's discuss there, in particular if you don't think it's an issue worth fixing.

henkbirkholz:
This kinda got old and we'll discuss it at the next editor's meeting.

henkbirkholz:
fixes #344

nedmsmith:
If tweaking this diagram, the "Collect Claims" arrow could be misinterpreted as Target sending claims to AE. If arrow text was "Collected Claims" then its less likely to be misinterpreted.

mcr:
I have added this diagram as a second digram, and moved the diagram to the end of section 3.

draft-ietf-rats-architecture.md
```diff
@@ -424,6 +424,17 @@ requires another component to feed various values to the TPM.
 Thus, an Attesting Environment in such a case would be the combination
 of the TPM together with whatever component is feeding it the measurements.
 
+## Environment Interactions
+
+As explained in the sections below, the Attesting Environment may be arbitrarily complex.
+It may be distributed in various ways, or may depend upon other environments that collect claims on the environment itself as part of the process.
```
dthaler, line 430: What does "distributed" mean here?   Distributing code for an attester?   Maybe pick another term?

other-attesting-env.txt
```diff
@@ -0,0 +1,27 @@
...
+|                   |     |                                   |
+|                   |     |                                   |
+|          Collect  |     |                                   | 
+|           Claims  |     |                                   |
+|                   |     |                                   |
+|                   v     |                                   |
+|                 .-------------.             .-------------. |
+|                 | Attesting   |             | Other       | |
+|                 | Environment |<------------| Attesting   | |
+|                 |             |   Evidence  | Environment | |
```
dthaler, line 24: Is this Evidence different from the Evidence in the vertical arrow?
Assuming so, then `Evidence'` or `Evidence 2` would be more clear, but either way the text discussing the diagram should explain what happens with this Evidence, since right now I don't understand this diagram.

dthaler (changes requested):

PR #351, opened 2021-09-02
Align the definition of Endorsement

### LINKED ISSUE #350: align endorsement definitions
In §4.2 Endorsement is defined as: _"A secure statement that an Endorser vouches for the integrity of an Attester's various capabilities such as Claims collection and Evidence signing."_

§8.2 uses a more restrictive description: _"An Endorsement is a secure statement that some entity (e.g., a manufacturer) vouches for the integrity of the device's signing capability."_, which seems to limit the scope of an Endorsement to the verification of the Attester's signature on Evidence.

I think §8.2 should be rephrased to align with §4.2 so to allow Endorsements to be more than just verification keys.

Besides, the definition of Endorser in §4.1: _"[…] Endorsements help Verifiers appraise the authenticity of Evidence"_ could be relaxed to drop the (seemingly) exclusive coupling of Endorsements with authenticity appraisal.

thomas-fossati:
* Align Section 8.2 with the definition in Section 4.2
* Relax the definition of Endorser in Section 4.1 to drop the exclusive
  coupling between Endorsements and authenticity appraisal

Fixes #350

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -587,7 +587,7 @@ Verifier Owner:
 
 Endorser:
 
-: A role performed by an entity (typically a manufacturer) whose Endorsements help Verifiers appraise the authenticity of Evidence.
+: A role performed by an entity (typically a manufacturer) whose Endorsements may help Verifiers appraise the authenticity of Evidence and infer further capabilities of the Attester.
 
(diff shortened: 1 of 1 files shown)


nedmsmith (commented):
I support these changes.

henkbirkholz (approved):
reads good to me

PR #373, opened 2021-11-23
clarify that Attestation Results can contain PII

### LINKED ISSUE #363: RD AD Attestation Result analogy
````
** Section 8.4.

   Thus, Attestation Results often need to include detailed information
   about the Attester, for use by Relying Parties, much like physical
   passports and drivers licenses include personal information such as
   name and date of birth.  
````
I'm not convinced this analogy is helpful.  Yes, it conveys the need for detailed information but it also immediately suggests Attestation Results might include PII or privacy sensitive identifiers.

### PR DESCRIPTION
**Pull request by mcr**
maybe close #363
unclear if this was what was intended

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -1063,9 +1063,8 @@ A more complex appraisal policy might involve using the information
 provided in the Attestation Result to compare against expected values, or to apply complex analysis
 of other information contained in the Attestation Result.
 
-Thus, Attestation Results often need to include detailed information about the Attester,
-for use by Relying Parties, much like physical passports and drivers licenses include
-personal information such as name and date of birth.  Unlike Evidence, which is often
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -1063,9 +1063,8 @@ A more complex appraisal policy might involve using the information
 provided in the Attestation Result to compare against expected values, or to apply complex analysis
 of other information contained in the Attestation Result.
 
-Thus, Attestation Results often need to include detailed information about the Attester,
-for use by Relying Parties, much like physical passports and drivers licenses include
-personal information such as name and date of birth.  Unlike Evidence, which is often
+Thus, Attestation Results can contain privacy sensitive information as discussed in section {{privacy-considerations}}.
```
henkbirkholz, line 1066:
```suggestion
Thus, Attestation Results can contain detailed information about an Attester, which can include privacy sensitive information as discussed in section {{privacy-considerations}}.
```

PR #428, opened 2022-07-28
try to capture comments/resolution to AD comments

mcr:
attempt to provide forward reference as requested on Monday

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -572,6 +572,9 @@ As explained in {{RFC4949}}, Section 2.6 when this document says "Compare:", the
 
 This document uses the following terms.
 
+As explained in {{overview}}, artifacts are defined by which role elements produce them or consume them.
+Other role elements may forward, pass-through, cache or repeat artifacts.
+
 ## Roles
 
 Attester:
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-architecture.md
```diff
@@ -572,6 +572,9 @@ As explained in {{RFC4949}}, Section 2.6 when this document says "Compare:", the
 
 This document uses the following terms.
 
+As explained in {{overview}}, artifacts are defined by which role elements produce them or consume them.
+Other role elements may forward, pass-through, cache or repeat artifacts.
```
thomas-fossati, line 576: what does "repeat" mean?  is it "make identical copies"?

dthaler, line 576: To Thomas's point, maybe:
```suggestion
As explained in {{overview}}, artifacts are defined by which roles produce them or consume them.
In some protocol instantiations, other roles may cache and forward artifacts as opaque data, to one or entities implementing the consuming role.
```

dthaler (approved):

dthaler:
This looks good to me, although I wonder whether the sentence would make more sense at the end of the "Implementation Considerations" section (a couple lines earlier in the document).  I am ok either way.

mcr:
> This looks good to me, although I wonder whether the sentence would make more sense at the end of the "Implementation Considerations" section (a couple lines earlier in the document). I am ok either way.

Works for me. That would put that into section 1.

thomas-fossati (approved):
LGTM

nedmsmith (commented):
"artifacts are defined by which roles produce them or consume them" - This sentence structure seems awkward. Is "artifacts are defined by the roles that produce or consume them" better?

PR #438, opened 2022-09-27
two additional typos found by SecDir review

### LINKED ISSUE #430: SecDir review: mitigation against contextual attack
The review is at:  https://mailarchive.ietf.org/arch/msg/last-call/0JpYvkgL-79nvmsrUpWKtMqtkhE
The review is positive, but asks for:
````
Even                       
if confidentiality is provided, the section goes on to state that information                        
can still be inferred by contextual or timing of the attestor exchange.  The                         
draft doesn’t describe ways to mitigate against this type of attack but should                       
give some guidance. 
````

### PR DESCRIPTION
**Pull request by mcr**
close #430

### DIFF
--- a/draft-ietf-rats-architecture.md
+++ b/draft-ietf-rats-architecture.md
@@ -243,7 +243,7 @@ Typically, a solution starts with a specific component (sometimes referred to as
 provides trustworthy device identity, and performs a series of operations that enables trustworthiness appraisals for other components.
 Such components perform operations that help determine the trustworthiness of yet other components,
 by collecting, protecting or signing measurements.
-Measurements that have been signed by such components comprise Evidence that when evaluated either supports or refutes a claim of trustworthiness.
+Measurements that have been signed by such components comprise of Evidence that when evaluated either supports or refutes a claim of trustworthiness.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: draft-ietf-rats-architecture.md [hunk 1]
```diff
@@ -243,7 +243,7 @@ Typically, a solution starts with a specific component (sometimes referred to as
 provides trustworthy device identity, and performs a series of operations that enables trustworthiness appraisals for other components.
 Such components perform operations that help determine the trustworthiness of yet other components,
 by collecting, protecting or signing measurements.
-Measurements that have been signed by such components comprise Evidence that when evaluated either supports or refutes a claim of trustworthiness.
+Measurements that have been signed by such components comprise of Evidence that when evaluated either supports or refutes a claim of trustworthiness.
```
dthaler, line 246: My reading of https://www.merriam-webster.com/dictionary/comprise is that this change is not grammatically correct.  It was ok as it was, but more readable would be
```suggestion
Measurements that have been signed by such components are comprised of Evidence that when evaluated either supports or refutes a claim of trustworthiness.
```
That is, "of" is used with "are comprised" but not "comprise".

dthaler (approved):

